PR #4, opened 2017-10-18
Update blacklisted objects

dmosorast:
Added all of the unqueryable objects found in my testing.

I have added EntityDefinition to the blanket blacklist, but with a note that it only threw on the RecordTypesSupported field. If we add support to remove specific fields from discovery (instead of sync, like attributes), this could be made more granular.

### DIFF
--- a/tap_salesforce/__init__.py
+++ b/tap_salesforce/__init__.py
@@ -23,6 +23,7 @@ CONFIG = {
 }
 
 BLACKLISTED_FIELDS = set(['attributes'])
+UNSUPPORTED_BULK_API_SALESFORCE_FIELDS = set([('RecordTypesSupported',"this field is unsupported by the Bulk API.")])
 
 # The following objects are not supported by the bulk API.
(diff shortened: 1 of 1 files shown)


ccapurso:
In terms of listing a `RecordTypesSupported` as unsupported, it might make sense to add the ability to determine which fields are unsupported based on map. Currently this is only done for compound fields and unsupported types. We could add the ability list arbitrary fields as unsupported using something like the following:

```
  {
    "object1": {
      "field1": "reason 1",
      "field2": "reason 2"
    }
  }
```

Using the lookup structure above, discovery mode would be able to determine that `inclusion` must be set to `unsupported` and that it should emit `unsupported-description` metadata based on the reason for a given field.

dmosorast:
@ccapurso, added field-level blacklisting and ran through a full sync. Ready to merge pending another set of eyes.

ccapurso:
Cool, this looks good.

PR #10, opened 2017-11-03
Write bookmarks for records within our date range; remove pendulum

KAllan357:


tap_salesforce/__init__.py
```diff
@@ -330,7 +331,10 @@ def do_sync(sf, catalog, state):
                                     record=rec,
                                     version=stream_version))
 
-                            if replication_key:
+                            # Before writing a bookmark, make sure Salesforce has not given us a record
+                            # with one outside our range
```
timvisher: Can we do two additional things here.

1. Add in the comment the situation we're trying to address. Someone coming to this tap 6 months from now will think this code is crazy unless we document what we saw.

1. log that it happened, probably with the context of the query range we expected, the actual value, and the position in the pagination.

    1. The query range gives us context for what we expected.

    1. The actual value gives us the detail of what we got back.

    1. The position in the pagination gives us some information about where, if any place, this kind of error tends to occur.

PR #12, opened 2017-11-21
handle byte, calculated, datacategorygroupreference, and masterrecord types

ccapurso:
We were missing the handling for a few types.

Primitive (https://developer.salesforce.com/docs/atlas.en-us.object_reference.meta/object_reference/primitive_data_types.htm):
- byte

Field Types (https://developer.salesforce.com/docs/atlas.en-us.object_reference.meta/object_reference/field_types.htm):
- calculated
- DataCategoryGroupReference (comes through as lowercase)
- masterrecord

### DIFF
--- a/tap_salesforce/salesforce/__init__.py
+++ b/tap_salesforce/salesforce/__init__.py
@@ -56,6 +56,14 @@ BINARY_TYPES = set([
     'byte'
 ])
 
+LOOSE_TYPES = set([
+    'anyType',
+
+    # A calculated field's type can be any of the supported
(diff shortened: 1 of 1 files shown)


tap_salesforce/salesforce/__init__.py
```diff
@@ -137,9 +145,13 @@ def field_to_property_schema(field, mdata):
         property_schema['type'] = "integer"
     elif sf_type == "time":
         property_schema['type'] = "string"
+    elif sf_type == "calculated":
```
KAllan357: Why not combine this with the `"anyType"` below?

KAllan357:
:+1:

PR #18, opened 2017-12-05
Use PK Chunking to execute Bulk Jobs

KAllan357:
When a bulk job experiences a QUERY_TIMEOUT error, try running it with the `Sforce-Enable-PKChunking` header.

Executing a job with that header will lead to more batches being created and a bookmark being emitted only at the end of processing.


tap_salesforce/salesforce/bulk.py
```diff
@@ -64,38 +68,76 @@ def _get_bulk_headers(self):
...
+    def _create_job(self, catalog_entry, pk_chunking=False):
         url = self.bulk_url.format(self.sf.instance_url, "job")
         body = {"operation": "queryAll", "object": catalog_entry['stream'], "contentType": "CSV"}
 
+        headers = self._get_bulk_headers()
+        headers['Sforce-Disable-Batch-Retry'] = "true"
+
+        if pk_chunking:
+            LOGGER.info("ADDING PK CHUNKING HEADER")
+            headers['Sforce-Enable-PKChunking'] = "true; chunkSize=50000"
```
KAllan357: Use the variable defined at the top.

tap_salesforce/salesforce/bulk.py
```diff
@@ -108,6 +150,24 @@ def _add_batch(self, catalog_entry, job_id, state):
 
         return batch['batchInfo']['id']
 
+    def _poll_on_pk_chunked_batch_status(self, job_id):
+        batches = self._get_batches(job_id)
+
+        completed_batches = []
+        failed_batches = []
```
KAllan357: Remove these

PR #19, opened 2017-12-13
Yield batch results correctly

KAllan357:


indigojump:
Approved pending tap-tester.

PR #20, opened 2017-12-13
Prevent exceptions during login from killing the thread

KAllan357:
Don't let an exception during the `POST` to kill the thread and prevent it from executing another login.

### DIFF
--- a/tap_salesforce/salesforce/__init__.py
+++ b/tap_salesforce/salesforce/__init__.py
@@ -276,6 +276,7 @@ class Salesforce(object):
 
         LOGGER.info("Attempting login via OAuth2")
 
+        resp = None
         try:
             resp = self.session.post(
                 login_url, data=login_body, headers={"Content-Type": "application/x-www-form-urlencoded"})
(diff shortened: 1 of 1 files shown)


tap_salesforce/salesforce/__init__.py
```diff
@@ -276,23 +276,23 @@ def login(self):
...
+            auth = resp.json()
 
-        self.access_token = auth['access_token']
-        self.instance_url = auth['instance_url']
-        self.login_timer = threading.Timer(REFRESH_TOKEN_EXPIRATION_PERIOD, self.login)
-        self.login_timer.start()
+            self.access_token = auth['access_token']
+            self.instance_url = auth['instance_url']
+        except Exception as e:
+            raise Exception(str(e) + ", Response from Salesforce: {}".format(resp.text)) from e
```
dmosorast: `resp.text` could possibly throw if session.post throws to get us into this. It's very useful information if it exists, maybe move it into an "if exists" check and add it on if so?

PR #42, opened 2018-03-12
Feature/use more metadata

psantacl:
replication key, replication method, key_properties, and table_key_properties into metadata


KAllan357:
👍

PR #56, opened 2018-10-17
Use should_sync_field from singer

iterati:
Requires updated singer-python (pending PR: singer-io/singer-python#79)


iterati:
@dmosorast I removed the integration tests

PR #57, opened 2018-11-01
Remove invalid job ids from state

dmosorast:
If a sync runs incompletely, and saves a job ID that it was waiting on, then the Job isn't requested for some time, it will be deleted from Salesforce.

This PR is to detect this case, save bookmarks as-is, remove the JobID/BatchIDs from the state, then move on.


tap_salesforce/__init__.py
```diff
@@ -307,12 +307,15 @@ def do_sync(sf, catalog, state):
                 LOGGER.info("%s: Completed sync (%s rows)", stream_name, counter.value)
                 state.get('bookmarks', {}).get(catalog_entry['tap_stream_id'], {}).pop('JobID', None)
```
KAllan357: Can we just add some comments here to explain what this is doing.

PR #58, opened 2018-11-01
Fix/background operation result full table

dmosorast:
The object BackgroundOperationResult does not support ordering by `CreatedDate`, so this must be requested in a `FULL_TABLE` manner.


KAllan357:
:+1:

PR #65, opened 2019-08-14
Add SOAP login support

cenxiao:
My most recent work requires me to pull data from a legacy Salesforce environment which doesn't have OAuth enabled, so I added the SOAP login support with the login code copied from Simple-Salesforce. I also added the two sets of required configs which looks pretty awkward, however, by looking at the original Singer.io documentation, I don't see this situation being mentioned.

### DIFF
--- a/tap_salesforce/__init__.py
+++ b/tap_salesforce/__init__.py
@@ -13,18 +13,31 @@ from tap_salesforce.salesforce.exceptions import (
 
 LOGGER = singer.get_logger()
 
-REQUIRED_CONFIG_KEYS = ['refresh_token',
-                        'client_id',
-                        'client_secret',
-                        'start_date',
-                        'api_type',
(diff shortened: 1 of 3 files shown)


cmerrick:
Hi @cenxiao, thanks for your contribution!

In order for us to evaluate and accept your PR, we ask that you [sign a contribution license agreement](https://stitch-cla-enforcer.herokuapp.com/). It's all electronic and will take just minutes.

cmerrick:
You did it @cenxiao!

Thank you for signing the Singer Contribution License Agreement.

ps2goat:
This one has been over a year! This would be a welcome change, so we don't have to go through getting a valid refresh token if something busted for more than a few days.

BenjMaq:
Honestly, it would be amazing if we got this merged.

PR #69, opened 2019-09-19
Add decimal to integer field coercion because the API is bad

luandy64:


luandy64:
The schema for the `DailyRequestTimeUsed` field on the `Site` stream changed from
```
          "DailyRequestTimeUsed": {
            "type": [
              "null",
              "integer"
            ]
          }
```
to

```
          "DailyRequestTimeUsed": {
            "type": [
              "null",
              "integer",
              "number"
            ]
          }
```

PR #70, opened 2019-10-08
Query bulk during discovery

nick-mccoy:
Since `api_type` is immutable, the tap should fail during discovery if the user selected `BULK` but doesn't have permissions.  Otherwise, the tap fails during sync mode, and since the user cannot change `api_type`, they have to recreate the connection.

The fix is to try and check the bulk quota usage at the beginning of discovery, which returns a 403 if the user doesn't have `BULK` permissions.

### DIFF
--- a/tap_salesforce/__init__.py
+++ b/tap_salesforce/__init__.py
@@ -1,7 +1,6 @@
 #!/usr/bin/env python3
 import json
 import sys
-import requests
 import singer
 import singer.utils as singer_utils
 from singer import metadata, metrics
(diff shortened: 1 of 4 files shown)


tap_salesforce/__init__.py
```diff
@@ -116,6 +117,18 @@ def do_discover(sf):
...
+    # Check if the user has BULK API enabled
+    if sf.api_type == 'BULK':
+        try:
+            bulk = Bulk(sf)
+            bulk.check_bulk_quota_usage()
+        except requests.exceptions.HTTPError as err:
+            if err.response is not None and len(err.response.json()) > 0:
+                for error_response_item in err.response.json():
+                    if error_response_item.get('errorCode') == 'API_DISABLED_FOR_ORG':
+                        raise TapSalesforceBulkAPIDisabledException('This client does not have Bulk API permissions, received "API_DISABLED_FOR_ORG" error code')
```
KAllan357: Perhaps this block belongs in the bulk module and could be "what happens" when you create an instance.

PR #86, opened 2020-08-25
added NetworkUserHistoryRecent to blacklist

jacobrobertbaca:
# Description of change
added NetworkUserHistoryRecent to blacklist
https://stitchdata.atlassian.net/secure/RapidBoard.jspa?rapidView=23&modal=detail&selectedIssue=SUP-1797

# QA steps
 - [ ] automated tests passing
- ran discovery with cloned connection of  client that was previously getting an error that had to do with object not being supported. stream was no longer written to catalog
 
# Risks
 - low
# Rollback steps
 - revert this branch

### DIFF
--- a/tap_salesforce/sync.py
+++ b/tap_salesforce/sync.py
@@ -103,8 +103,8 @@ def sync_stream(sf, catalog_entry, state):
             sync_records(sf, catalog_entry, state, counter)
             singer.write_state(state)
         except RequestException as ex:
-            raise Exception("Error syncing {}: {} Response: {}".format(# pylint: disable=raise-missing-from
-                stream, ex, ex.response.text))
+            raise Exception("Error syncing {}: {} Response: {}".format(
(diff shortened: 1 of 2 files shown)


tap_salesforce/sync.py
```diff
@@ -103,7 +103,7 @@ def sync_stream(sf, catalog_entry, state):
             sync_records(sf, catalog_entry, state, counter)
             singer.write_state(state)
         except RequestException as ex:
-            raise Exception("Error syncing {}: {} Response: {}".format(
+            raise Exception("Error syncing {}: {} Response: {}".format(# pylint: disable=raise-missing-from
```
dmosorast, line 106: If you're going to make a change here, the better action would probably be to add the `from` to it. like `raise Exception(...) from ex`

I think this was written before we did pylint and before we understood Python better. The `from` syntax makes the stack trace more readable.

PR #135, opened 2022-03-31
TDL-16115: Add lookback window advanced config option for Incremental Syncs

harshpatel4crest:
# Description of change
TDL-16115: Add lookback window advanced config option for Incremental Syncs
- Added configurable lookback window with the default value of 10 seconds.
- Added integration test `test_salesforce_lookback_window` for validating lookback window

# QA steps
 - [ ] automated tests passing
 
# Risks

# Rollback steps
 - revert this branch

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@
 [![CircleCI Build Status](https://circleci.com/gh/singer-io/tap-salesforce.png)](https://circleci.com/gh/singer-io/tap-salesforce.png)
 
 
-[Singer](https://www.singer.io/) tap that extracts data from a [Salesforce](https://www.salesforce.com/) database and produces JSON-formatted data following the [Singer spec](https://github.com/singer-io/getting-started/blob/master/SPEC.md).
+[Singer](https://www.singer.io/) tap that extracts data from a [Salesforce](https://www.salesforce.com/) database and produces JSON-formatted data following the [Singer spec](https://github.com/singer-io/getting-started/blob/master/docs/SPEC.md#singer-specification).
 
 ```bash
 $ mkvirtualenv -p python3 tap-salesforce
--- a/README.md
+++ b/README.md
@@ -31,6 +31,7 @@ $ tap-salesforce --config config.json --properties properties.json --state state
   "start_date": "2017-11-02T00:00:00Z",
   "api_type": "BULK",
   "select_fields_by_default": true
+  "lookback_window": 10
 }
 ```
 
@@ -39,6 +40,8 @@ The `client_id` and `client_secret` keys are your OAuth Salesforce App secrets. 
 The `start_date` is used by the tap as a bound on SOQL queries when searching for records.  This should be an [RFC3339](https://www.ietf.org/rfc/rfc3339.txt) formatted date-time, like "2018-01-08T00:00:00Z". For more details, see the [Singer best practices for dates](https://github.com/singer-io/getting-started/blob/master/BEST_PRACTICES.md#dates).
 
 The `api_type` is used to switch the behavior of the tap between using Salesforce's "REST" and "BULK" APIs. When new fields are discovered in Salesforce objects, the `select_fields_by_default` key describes whether or not the tap will select those fields by default.
+
+The `lookback_window` (in seconds) subtracts the desired amount of seconds from the start date to sync past data. Default value: 10 seconds.
 
 ## Run Discovery
 
--- a/README.md
+++ b/README.md
@@ -41,7 +41,7 @@ The `start_date` is used by the tap as a bound on SOQL queries when searching fo
 
 The `api_type` is used to switch the behavior of the tap between using Salesforce's "REST" and "BULK" APIs. When new fields are discovered in Salesforce objects, the `select_fields_by_default` key describes whether or not the tap will select those fields by default.
 
-The `lookback_window` (in seconds) subtracts the desired amount of seconds from the bookmark to sync past data. Default value: 10 seconds.
+The `lookback_window` (in seconds) subtracts the desired amount of seconds from the bookmark to sync past data.
 
 ## Run Discovery
 
--- a/README.md
+++ b/README.md
@@ -41,7 +41,7 @@ The `start_date` is used by the tap as a bound on SOQL queries when searching fo
 
 The `api_type` is used to switch the behavior of the tap between using Salesforce's "REST" and "BULK" APIs. When new fields are discovered in Salesforce objects, the `select_fields_by_default` key describes whether or not the tap will select those fields by default.
 
-The `lookback_window` (in seconds) subtracts the desired amount of seconds from the start date to sync past data. Default value: 10 seconds.
+The `lookback_window` (in seconds) subtracts the desired amount of seconds from the bookmark to sync past data. Default value: 10 seconds.
 
 ## Run Discovery
 
--- a/README.md
+++ b/README.md
@@ -41,7 +41,7 @@ The `start_date` is used by the tap as a bound on SOQL queries when searching fo
 
 The `api_type` is used to switch the behavior of the tap between using Salesforce's "REST" and "BULK" APIs. When new fields are discovered in Salesforce objects, the `select_fields_by_default` key describes whether or not the tap will select those fields by default.
 
-The `lookback_window` (in seconds) subtracts the desired amount of seconds from the bookmark to sync past data.
+The `lookback_window` (in seconds) subtracts the desired amount of seconds from the bookmark to sync past data. Recommended value: 10 seconds.
 
 ## Run Discovery
 
--- a/tap_salesforce/__init__.py
+++ b/tap_salesforce/__init__.py
@@ -6,7 +6,7 @@ import singer.utils as singer_utils
 from singer import metadata, metrics
 import tap_salesforce.salesforce
 from tap_salesforce.sync import (sync_stream, resume_syncing_bulk_query, get_stream_version)
-from tap_salesforce.salesforce import Salesforce, DEFAULT_LOOKBACK_WINDOW
+from tap_salesforce.salesforce import Salesforce
 from tap_salesforce.salesforce.bulk import Bulk
 from tap_salesforce.salesforce.exceptions import (
     TapSalesforceException, TapSalesforceQuotaExceededException, TapSalesforceBulkAPIDisabledException)
@@ -375,6 +375,10 @@ def main_impl():
 
     sf = None
     try:
+        # get lookback window from config
+        lookback_window = CONFIG.get('lookback_window')
+        lookback_window = int(lookback_window) if lookback_window else None
+
         sf = Salesforce(
             refresh_token=CONFIG['refresh_token'],
             sf_client_id=CONFIG['client_id'],
@@ -385,7 +389,7 @@ def main_impl():
             select_fields_by_default=CONFIG.get('select_fields_by_default'),
             default_start_date=CONFIG.get('start_date'),
             api_type=CONFIG.get('api_type'),
-            lookback_window=int(CONFIG.get('lookback_window', DEFAULT_LOOKBACK_WINDOW)))
+            lookback_window=lookback_window)
         sf.login()
 
         if args.discover:
--- a/tap_salesforce/__init__.py
+++ b/tap_salesforce/__init__.py
@@ -1,8 +1,8 @@
 #!/usr/bin/env python3
 import datetime
-from dateutil.parser import parse
 import json
 import sys
+from dateutil.parser import parse
 import singer
 import singer.utils as singer_utils
 from singer import metadata, metrics
--- a/tap_salesforce/__init__.py
+++ b/tap_salesforce/__init__.py
@@ -6,7 +6,7 @@ import singer.utils as singer_utils
 from singer import metadata, metrics
 import tap_salesforce.salesforce
 from tap_salesforce.sync import (sync_stream, resume_syncing_bulk_query, get_stream_version)
-from tap_salesforce.salesforce import Salesforce
+from tap_salesforce.salesforce import Salesforce, DEFAULT_LOOKBACK_WINDOW
 from tap_salesforce.salesforce.bulk import Bulk
 from tap_salesforce.salesforce.exceptions import (
     TapSalesforceException, TapSalesforceQuotaExceededException, TapSalesforceBulkAPIDisabledException)
@@ -385,7 +385,7 @@ def main_impl():
             select_fields_by_default=CONFIG.get('select_fields_by_default'),
             default_start_date=CONFIG.get('start_date'),
             api_type=CONFIG.get('api_type'),
-            lookback_window=int(CONFIG.get('lookback_window', 10)))
+            lookback_window=int(CONFIG.get('lookback_window', DEFAULT_LOOKBACK_WINDOW)))
         sf.login()
 
         if args.discover:
--- a/tap_salesforce/__init__.py
+++ b/tap_salesforce/__init__.py
@@ -1,8 +1,6 @@
 #!/usr/bin/env python3
-import datetime
 import json
 import sys
-from dateutil.parser import parse
 import singer
 import singer.utils as singer_utils
 from singer import metadata, metrics
@@ -75,10 +73,9 @@ def build_state(raw_state, catalog):
             job_id = singer.get_bookmark(raw_state, tap_stream_id, 'JobID')
             batches = singer.get_bookmark(raw_state, tap_stream_id, 'BatchIDs')
             current_bookmark = singer.get_bookmark(raw_state, tap_stream_id, 'JobHighestBookmarkSeen')
-            adjusted_current_bookmark = singer_utils.strftime(singer_utils.strptime_with_tz(current_bookmark) - datetime.timedelta(seconds=int(CONFIG.get('lookback_window', 10))))
             state = singer.write_bookmark(state, tap_stream_id, 'JobID', job_id)
             state = singer.write_bookmark(state, tap_stream_id, 'BatchIDs', batches)
-            state = singer.write_bookmark(state, tap_stream_id, 'JobHighestBookmarkSeen', adjusted_current_bookmark)
+            state = singer.write_bookmark(state, tap_stream_id, 'JobHighestBookmarkSeen', current_bookmark)
 
         if replication_method == 'INCREMENTAL':
             replication_key = catalog_metadata.get((), {}).get('replication-key')
@@ -89,9 +86,8 @@ def build_state(raw_state, catalog):
                 state = singer.write_bookmark(
                     state, tap_stream_id, 'version', version)
             if replication_key_value is not None:
-                adjusted_replication_key_value = singer_utils.strftime(singer_utils.strptime_with_tz(replication_key_value) - datetime.timedelta(seconds=int(CONFIG.get('lookback_window', 10))))
                 state = singer.write_bookmark(
-                    state, tap_stream_id, replication_key, adjusted_replication_key_value)
+                    state, tap_stream_id, replication_key, replication_key_value)
         elif replication_method == 'FULL_TABLE' and version is None:
             state = singer.write_bookmark(state, tap_stream_id, 'version', version)
 
@@ -376,8 +372,6 @@ def do_sync(sf, catalog, state):
 def main_impl():
     args = singer_utils.parse_args(REQUIRED_CONFIG_KEYS)
     CONFIG.update(args.config)
-    adjusted_start_date = singer_utils.strftime(singer_utils.strptime_with_tz(CONFIG['start_date']) - datetime.timedelta(seconds=int(CONFIG.get('lookback_window', 10))))
-    adjusted_start_date = parse(adjusted_start_date).strftime("%Y-%m-%dT%H:%M:%SZ")
 
     sf = None
     try:
@@ -389,8 +383,9 @@ def main_impl():
             quota_percent_per_run=CONFIG.get('quota_percent_per_run'),
             is_sandbox=CONFIG.get('is_sandbox'),
             select_fields_by_default=CONFIG.get('select_fields_by_default'),
-            default_start_date=adjusted_start_date,
-            api_type=CONFIG.get('api_type'))
+            default_start_date=CONFIG.get('start_date'),
+            api_type=CONFIG.get('api_type'),
+            lookback_window=int(CONFIG.get('lookback_window', 10)))
         sf.login()
 
         if args.discover:
--- a/tap_salesforce/__init__.py
+++ b/tap_salesforce/__init__.py
@@ -1,4 +1,6 @@
 #!/usr/bin/env python3
+import datetime
+from dateutil.parser import parse
 import json
 import sys
 import singer
@@ -73,9 +75,10 @@ def build_state(raw_state, catalog):
             job_id = singer.get_bookmark(raw_state, tap_stream_id, 'JobID')
             batches = singer.get_bookmark(raw_state, tap_stream_id, 'BatchIDs')
             current_bookmark = singer.get_bookmark(raw_state, tap_stream_id, 'JobHighestBookmarkSeen')
+            adjusted_current_bookmark = singer_utils.strftime(singer_utils.strptime_with_tz(current_bookmark) - datetime.timedelta(seconds=int(CONFIG.get('lookback_window', 10))))
             state = singer.write_bookmark(state, tap_stream_id, 'JobID', job_id)
             state = singer.write_bookmark(state, tap_stream_id, 'BatchIDs', batches)
-            state = singer.write_bookmark(state, tap_stream_id, 'JobHighestBookmarkSeen', current_bookmark)
+            state = singer.write_bookmark(state, tap_stream_id, 'JobHighestBookmarkSeen', adjusted_current_bookmark)
 
         if replication_method == 'INCREMENTAL':
             replication_key = catalog_metadata.get((), {}).get('replication-key')
@@ -86,8 +89,9 @@ def build_state(raw_state, catalog):
                 state = singer.write_bookmark(
                     state, tap_stream_id, 'version', version)
             if replication_key_value is not None:
+                adjusted_replication_key_value = singer_utils.strftime(singer_utils.strptime_with_tz(replication_key_value) - datetime.timedelta(seconds=int(CONFIG.get('lookback_window', 10))))
                 state = singer.write_bookmark(
-                    state, tap_stream_id, replication_key, replication_key_value)
+                    state, tap_stream_id, replication_key, adjusted_replication_key_value)
         elif replication_method == 'FULL_TABLE' and version is None:
             state = singer.write_bookmark(state, tap_stream_id, 'version', version)
 
@@ -372,6 +376,8 @@ def do_sync(sf, catalog, state):
 def main_impl():
     args = singer_utils.parse_args(REQUIRED_CONFIG_KEYS)
     CONFIG.update(args.config)
+    adjusted_start_date = singer_utils.strftime(singer_utils.strptime_with_tz(CONFIG['start_date']) - datetime.timedelta(seconds=int(CONFIG.get('lookback_window', 10))))
+    adjusted_start_date = parse(adjusted_start_date).strftime("%Y-%m-%dT%H:%M:%SZ")
 
     sf = None
     try:
@@ -383,7 +389,7 @@ def main_impl():
             quota_percent_per_run=CONFIG.get('quota_percent_per_run'),
             is_sandbox=CONFIG.get('is_sandbox'),
             select_fields_by_default=CONFIG.get('select_fields_by_default'),
-            default_start_date=CONFIG.get('start_date'),
+            default_start_date=adjusted_start_date,
             api_type=CONFIG.get('api_type'))
         sf.login()
 
--- a/tap_salesforce/__init__.py
+++ b/tap_salesforce/__init__.py
@@ -422,5 +422,6 @@ def main():
         LOGGER.critical(e)
         sys.exit(1)
     except Exception as e:
-        LOGGER.critical(e)
+        for error_line in str(e).splitlines():
+            LOGGER.critical(error_line)
         raise e
--- a/tap_salesforce/salesforce/__init__.py
+++ b/tap_salesforce/salesforce/__init__.py
@@ -17,9 +17,6 @@ from tap_salesforce.salesforce.exceptions import (
 
 LOGGER = singer.get_logger()
 
-# default lookback window of 10 seconds
-DEFAULT_LOOKBACK_WINDOW = 10
-
 # The minimum expiration setting for SF Refresh Tokens is 15 minutes
 REFRESH_TOKEN_EXPIRATION_PERIOD = 900
 
@@ -215,7 +212,7 @@ class Salesforce():
                  select_fields_by_default=None,
                  default_start_date=None,
                  api_type=None,
-                 lookback_window=DEFAULT_LOOKBACK_WINDOW):
+                 lookback_window=None):
         self.api_type = api_type.upper() if api_type else None
         self.refresh_token = refresh_token
         self.token = token
@@ -243,7 +240,7 @@ class Salesforce():
         self.lookback_window = lookback_window
 
         # validate start_date
-        singer_utils.strptime(default_start_date)
+        singer_utils.strptime_to_utc(default_start_date)
 
     def _get_standard_headers(self):
         return {"Authorization": "Bearer {}".format(self.access_token)}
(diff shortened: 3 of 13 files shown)


tap_salesforce/__init__.py
```diff
@@ -86,8 +89,9 @@ def build_state(raw_state, catalog):
                 state = singer.write_bookmark(
                     state, tap_stream_id, 'version', version)
             if replication_key_value is not None:
+                adjusted_replication_key_value = singer_utils.strftime(singer_utils.strptime_with_tz(replication_key_value) - datetime.timedelta(seconds=int(CONFIG.get('lookback_window', 10))))
```
dbshah1212, line 92: Why do we need to define it again?

tap_salesforce/__init__.py
```diff
@@ -73,9 +75,10 @@ def build_state(raw_state, catalog):
             job_id = singer.get_bookmark(raw_state, tap_stream_id, 'JobID')
             batches = singer.get_bookmark(raw_state, tap_stream_id, 'BatchIDs')
             current_bookmark = singer.get_bookmark(raw_state, tap_stream_id, 'JobHighestBookmarkSeen')
+            adjusted_current_bookmark = singer_utils.strftime(singer_utils.strptime_with_tz(current_bookmark) - datetime.timedelta(seconds=int(CONFIG.get('lookback_window', 10))))
```
dbshah1212, line 78: Can we make a small function of it? It will increase the readability.

harshpatel4crest: Updated the code.

tap_salesforce/salesforce/__init__.py
```diff
@@ -381,13 +384,16 @@ def _get_selected_properties(self, catalog_entry):
...
 
-        return (singer.get_bookmark(state,
-                                    catalog_entry['tap_stream_id'],
-                                    replication_key) or self.default_start_date)
+        start_date = singer.get_bookmark(state, catalog_entry['tap_stream_id'], replication_key) or self.default_start_date
+        if not with_lookback:
+            return start_date
+
+        adjusted_start_date = singer_utils.strftime(singer_utils.strptime_with_tz(start_date) - datetime.timedelta(seconds=self.lookback_window))
+        return adjusted_start_date
```
savan-chovatiya, line 396: Add code comment here.

tap_salesforce/salesforce/__init__.py
```diff
@@ -381,13 +384,16 @@ def _get_selected_properties(self, catalog_entry):
                                             self.select_fields_by_default)]
 
 
-    def get_start_date(self, state, catalog_entry):
+    def get_start_date(self, state, catalog_entry, with_lookback=True):
```
savan-chovatiya, line 387: Change the default value of `with_lookback` to False and explicitly pass True as an argument wherever required. it will give more clarity regarding where lookback is used rather than not used.

harshpatel4crest: Updated.

tap_salesforce/salesforce/__init__.py
harshpatel4crest: Added comments.

README.md
```diff
@@ -40,6 +41,8 @@ The `start_date` is used by the tap as a bound on SOQL queries when searching fo
 
 The `api_type` is used to switch the behavior of the tap between using Salesforce's "REST" and "BULK" APIs. When new fields are discovered in Salesforce objects, the `select_fields_by_default` key describes whether or not the tap will select those fields by default.
 
+The `lookback_window` (in seconds) subtracts the desired amount of seconds from the start date to sync past data. Default value: 10 seconds.
```
savan-chovatiya, line 44:
```suggestion
The `lookback_window` (in seconds) subtracts the desired amount of seconds from the bookmark to sync past data. Default value: 10 seconds.
```

tap_salesforce/salesforce/__init__.py
```diff
@@ -381,13 +384,20 @@ def _get_selected_properties(self, catalog_entry):
...
-    def get_start_date(self, state, catalog_entry):
+    def get_start_date(self, state, catalog_entry, without_lookback=True):
         catalog_metadata = metadata.to_map(catalog_entry['metadata'])
         replication_key = catalog_metadata.get((), {}).get('replication-key')
 
-        return (singer.get_bookmark(state,
-                                    catalog_entry['tap_stream_id'],
-                                    replication_key) or self.default_start_date)
+        sync_start_date = singer.get_bookmark(state, catalog_entry['tap_stream_id'], replication_key) or self.default_start_date
+        # return date from the state file or start date if, 'without_lookback' is True
```
savan-chovatiya, line 392:
```suggestion
        # return bookmark from the state or start date if 'without_lookback' is True
```

tap_salesforce/salesforce/__init__.py
```diff
@@ -381,13 +384,20 @@ def _get_selected_properties(self, catalog_entry):
...
+        if without_lookback:
+            return sync_start_date
+
+        # if state file contains bookmark, subtract lookback window from the 'sync_start_date'
```
savan-chovatiya, line 396:
```suggestion
        # if the state contains a bookmark, subtract the lookback window from the bookmark
```

savan-chovatiya (changes requested):

tap_salesforce/salesforce/__init__.py
harshpatel4crest: Updated.

tap_salesforce/salesforce/__init__.py
harshpatel4crest: Updated.

README.md
harshpatel4crest: Updated the file.

savan-chovatiya (approved):

tap_salesforce/salesforce/__init__.py
```diff
@@ -381,13 +384,20 @@ def _get_selected_properties(self, catalog_entry):
                                             self.select_fields_by_default)]
 
 
-    def get_start_date(self, state, catalog_entry):
+    def get_start_date(self, state, catalog_entry, without_lookback=True):
```
dbshah1212, line 387: Add code comment and explain the effect of the without_lookback variable.

harshpatel4crest: Added function comment

dbshah1212 (approved):

harshpatel4crest:
The Integration tests are failing because we need to add `lookback_window` config param in the UI side.

tests/test_salesforce_lookback_window.py
```diff
@@ -0,0 +1,108 @@
+from datetime import datetime, timedelta
+from tap_tester import connections, runner, menagerie
+from base import SalesforceBaseTest
+
+class SalesforceLookbackWindow(SalesforceBaseTest):
+
+    # subtract the desired amount of seconds form the date and return
+    def timedelta_formatted(self, dtime, seconds=0):
```
kspeer825, line 8: It would be better if this method took in the format as a parameter so that it's usage is explicit. We should never manipulate results in the tests without being explicit, else we may cover up a bug.

tests/test_salesforce_lookback_window.py
```diff
@@ -0,0 +1,108 @@
...
+        # get synced records
+        sync_records = runner.get_records_from_target_output()
+
+        # get replication keys
+        expected_replication_keys = self.expected_replication_keys()
+
+        # get start date ie. date from which the sync started
+        start_date = state.get('bookmarks').get('Account').get('SystemModstamp')
+        # calculate the simulated start date by subtracting lookback window seconds
+        start_date_with_lookback_window = self.timedelta_formatted(start_date, seconds=-self.get_properties()['lookback_window'])
```
kspeer825, line 82: Can you change the wording so that we do not mention start date? That term should be reserved for historical syncs. The point at which the sync starts on an incremental sync is based on the state you have set above, not the start date. Maybe I'm just being picky but I think it could be confusing to a tester in the future.

tests/test_salesforce_lookback_window.py
```diff
@@ -0,0 +1,108 @@
...
+    def name():
+        return 'tap_tester_salesforce_lookback_window'
+
+    @staticmethod
+    def get_properties():  # pylint: disable=arguments-differ
+        return {
+            'start_date' : '2021-11-10T00:00:00Z',
+            'instance_url': 'https://singer2-dev-ed.my.salesforce.com',
+            'select_fields_by_default': 'true',
+            'api_type': 'REST',
```
kspeer825, line 35: Can this test be executed against the BULK API as well?

kspeer825 (changes requested):

tests/test_salesforce_lookback_window.py
harshpatel4crest: Added format param in the function.

tests/test_salesforce_lookback_window.py
harshpatel4crest: Updated the variable name.

tests/test_salesforce_lookback_window.py
harshpatel4crest: Updated the code to run for BULK API too.

luandy64 (changes requested):
Just to reiterate: I think the current implementation is confusing because of the way things are named and it is inconsistent with other taps where we've implemented a lookback window.

I would like to hear some reasons to defend this implementation before I am comfortable approving this PR.

---

For the tests, I think we should be careful with adding too much extra code. 

For unit tests especially - if we are only trying to focus on 

```python
sf.get_start_date({}, mock_catalog_entry, without_lookback=False)
```

versus

```python
sf.get_start_date(mock_state, mock_catalog_entry, without_lookback=False)
```

Then we should write the bare minimum amount of code to allow us to call `get_start_date()`. In this case in particular, we don't need readers to spend time looking at how a `Salesforce` object is created. We need them to focus on 
1. A `Salesforce` object is created with a start date
2. This is what the catalog entry looks like
3. This is what the state looks like
4. Here is what `get_start_date` returns

tests/test_salesforce_lookback_window.py
```diff
@@ -0,0 +1,105 @@
...
+
+                    # Verify the sync records respect the (simulated) bookmark value
+                    self.assertGreaterEqual(self.parse_date(replication_key_value), self.parse_date(bookmark_with_lookback_window),
+                                            msg='The record does not respect the lookback window.')
+
+                    # verify if the record's bookmark value is between bookmark and (simulated) bookmark value
+                    if self.parse_date(bookmark_with_lookback_window) <= self.parse_date(replication_key_value) < self.parse_date(bookmark):
+                        is_between = True
+
+                    self.assertTrue(is_between, msg='No record found between bookmark and lookback date.')
```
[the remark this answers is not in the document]
harshpatel4crest: Ideally, if we are not using any lookback then we will get all data greater than the bookmark. Here, we are using lookback so, above assertion to verify we getting at least some data between `bookmark - lookback` and `bookmark`

tests/unittests/test_lookback_window.py
```diff
@@ -0,0 +1,362 @@
...
+        """
+            Test case to verify default lookback window (10 seconds) is passed if user has not passed from the config.
+        """
+
+        tap_salesforce.CONFIG = {
+            'refresh_token': None,
+            'client_id': None,
+            'client_secret': None,
+            'start_date': None
+        }
```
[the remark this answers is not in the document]
harshpatel4crest: As CONFIG is a global variable, and the tap is updating the variable [here](https://github.com/singer-io/tap-salesforce/blob/master/tap_salesforce/__init__.py#L374). So, when we will run the tests, the CONFIG will contain the previous values from the test (as shown in the image). Thus, we are resetting the value.

image

tap_salesforce/salesforce/__init__.py
```diff
@@ -381,13 +384,24 @@ def _get_selected_properties(self, catalog_entry):
...
```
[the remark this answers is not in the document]
harshpatel4crest: The Tap is using `get_start_date` in `sync.py` for bookmarking so, at that time, we do not want our bookmark to go backward. Thus, here we are getting the date without the lookback window therefore we have added an argument `without_lookback`. But to avoid confusion we have renamed that argument to `with_lookback`.

In the DoD of the card [TDL-16115](https://jira.talendforge.org/browse/TDL-16115). There is mentioned we need to add a default lookback of 10 seconds. So, we kept the default value of 10 rather than 0.

tap_salesforce/salesforce/__init__.py
```diff
@@ -381,13 +384,24 @@ def _get_selected_properties(self, catalog_entry):
...
+        """
+            if 'without_lookback' is True, then return start date or state file date
+            else subtract lookback window from the start date or state file date and return
```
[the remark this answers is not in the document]
harshpatel4crest: Right. Updated the comment.

tap_salesforce/__init__.py
```diff
@@ -384,7 +384,8 @@ def main_impl():
             is_sandbox=CONFIG.get('is_sandbox'),
             select_fields_by_default=CONFIG.get('select_fields_by_default'),
             default_start_date=CONFIG.get('start_date'),
-            api_type=CONFIG.get('api_type'))
+            api_type=CONFIG.get('api_type'),
+            lookback_window=int(CONFIG.get('lookback_window', 10)))
```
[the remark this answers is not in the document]
harshpatel4crest: Imported the variable containing the default lookback window size from the **tap_salesforce/salesforce/__init__.py** file.

tap_salesforce/salesforce/__init__.py
```diff
@@ -210,7 +211,8 @@ def __init__(self,
                  is_sandbox=None,
                  select_fields_by_default=None,
                  default_start_date=None,
-                 api_type=None):
+                 api_type=None,
+                 lookback_window=10):
```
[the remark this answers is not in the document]
harshpatel4crest: Added a global variable `DEFAULT_LOOKBACK_WINDOW`.

tests/unittests/test_lookback_window.py
```diff
@@ -0,0 +1,362 @@
...
+from unittest import mock
+import tap_salesforce
+
+# mock 'Salesforce' class
+class MockSalesforce:
+    rest_requests_attempted = 0
+    jobs_completed = 0
+    login_timer = None
+    def __init__(self, *args, **kwargs):
+        return None
```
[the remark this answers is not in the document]
harshpatel4crest: Updated.

tests/unittests/test_lookback_window.py
```diff
@@ -0,0 +1,362 @@
...
+
+        # function call with apply lookback as 'False'
+        start_date = sf.get_start_date({}, mock_catalog_entry)
+
+        # verify 20 seconds were subtracted from start date
+        self.assertEqual(start_date, '2021-01-02T00:00:00Z')
+
+    def test_no_lookback_window_subtraction_with_state__get_start_date(self):
+        """
+            Test case to verify state file date is not changed when we pass 'with_lookback=False'
```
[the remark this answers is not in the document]
harshpatel4crest: Updated the code.

tests/unittests/test_lookback_window.py
```diff
@@ -0,0 +1,362 @@
...
+
+        # function call with apply lookback as 'True'
+        start_date = sf.get_start_date(mock_state, mock_catalog_entry, without_lookback=False)
+
+        # verify 20 seconds were subtracted from state file date
+        self.assertEqual(start_date, '2021-01-09T23:59:40.000000Z')
+
+    def test_no_lookback_window_subtraction__get_start_date(self):
+        """
+            Test case to verify start date is not changed when we pass 'with_lookback=False'
```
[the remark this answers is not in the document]
harshpatel4crest: Updated the code.

tests/unittests/test_lookback_window.py
```diff
@@ -0,0 +1,362 @@
...
+
+        # create Salesforce object
+        sf = Salesforce(
+            refresh_token='test_refresh_token',
+            sf_client_id='test_client_id',
+            sf_client_secret='test_client_secret',
+            default_start_date=config.get('start_date'),
+            lookback_window=int(config.get('lookback_window', 10)))
+
+        # function call with apply lookback as 'True'
```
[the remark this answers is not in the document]
harshpatel4crest: Updated the tests code and variable name in the Tap.

tests/unittests/test_lookback_window.py
```diff
@@ -0,0 +1,362 @@
...
+            'refresh_token': 'test_refresh_token',
+            'client_id': 'test_client_id',
+            'client_secret': 'test_client_secret',
+            'start_date': '2021-01-02T00:00:00Z',
+            'api_type': 'REST',
+            'select_fields_by_default': 'true',
+            'lookback_window': 20
+        }
+        # mock parse args
+        mocked_parse_args.return_value = get_args(mock_config)
```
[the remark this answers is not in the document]
harshpatel4crest: Updated as per suggested.

tests/unittests/test_lookback_window.py
```diff
@@ -0,0 +1,362 @@
...
+
+    def login(self):
+        pass
+
+# mock args and return desired state, catalog and config file
+class MockParseArgs:
+    config = {}
+    discover = None
+    properties = None
+    state = None
```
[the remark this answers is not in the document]
harshpatel4crest: Removed the variables.

tests/test_salesforce_lookback_window.py
```diff
@@ -0,0 +1,105 @@
+from datetime import datetime, timedelta
+from tap_tester import connections, runner, menagerie
+from base import SalesforceBaseTest
+
+class SalesforceLookbackWindow(SalesforceBaseTest):
+
+    # subtract the desired amount of seconds form the date and return
+    def timedelta_formatted(self, dtime, format, seconds=0):
+        date_stripped = datetime.strptime(dtime, format)
+        return_date = date_stripped + timedelta(seconds=seconds)
```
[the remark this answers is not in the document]
harshpatel4crest: Updated the function name and updated the mechanism of the function to subtract seconds from the date.

tests/unittests/test_lookback_window.py
```diff
@@ -0,0 +1,362 @@
...
+            ],
+        }
+
+        # create Salesforce object
+        sf = Salesforce(
+            refresh_token='test_refresh_token',
+            sf_client_id='test_client_id',
+            sf_client_secret='test_client_secret',
+            default_start_date=config.get('start_date'),
+            lookback_window=int(config.get('lookback_window', 10)))
```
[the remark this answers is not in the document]
harshpatel4crest: Updated as per suggestion.

tests/unittests/test_lookback_window.py
```diff
@@ -0,0 +1,362 @@
...
+            ],
+        }
+
+        # create 'Salesforce' object
+        sf = Salesforce(
+            refresh_token='test_refresh_token',
+            sf_client_id='test_client_id',
+            sf_client_secret='test_client_secret',
+            default_start_date=config.get('start_date'),
+            lookback_window=int(config.get('lookback_window', 10)))
```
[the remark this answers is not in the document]
harshpatel4crest: Updated as per suggestion.

tests/unittests/test_lookback_window.py
```diff
@@ -0,0 +1,362 @@
...
+            }
+        }
+
+        # create 'Salesforce' object
+        sf = Salesforce(
+            refresh_token='test_refresh_token',
+            sf_client_id='test_client_id',
+            sf_client_secret='test_client_secret',
+            default_start_date=config.get('start_date'),
+            lookback_window=int(config.get('lookback_window', 10)))
```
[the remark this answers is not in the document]
harshpatel4crest: Updated as per suggestion.

tests/unittests/test_lookback_window.py
```diff
@@ -0,0 +1,362 @@
...
```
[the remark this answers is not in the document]
harshpatel4crest: Updated as per suggestion.

tests/unittests/test_lookback_window.py
```diff
@@ -0,0 +1,362 @@
...
```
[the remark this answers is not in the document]
harshpatel4crest: Updated as per suggestion.

tests/unittests/test_lookback_window.py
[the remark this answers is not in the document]
luandy64: But you already mocked the value of `args.config` in the test
https://github.com/singer-io/tap-salesforce/blob/cc4eb605190f697c5330ef2b4dfa31782635e7fb/tests/unittests/test_lookback_window.py#L126
https://github.com/singer-io/tap-salesforce/blob/d89a18f4f4f95018cb508c8aea5522e74ea35b98/tap_salesforce/__init__.py#L373-L374

So I agree with what you are saying: the first test will have the `CONFIG` with `None`s in the values and every test after that will have previous values.

But in this case it doesn't matter because your test is mocking the value before `tap_salesforce.main_impl()` is called. So to reset the config to `{'refresh_token': None, ...}` doesn't need to happen since you immediately reset it to `get_args()`

---

This is what the code looks like now

```
tap-salesforce config   {'refresh_token': None, ...}

           First test   SET {'refresh_token': None, ...}
           First test   SET {'refresh_token': 'test_refresh_token', ...}
           First test   RUN test

          Second test   SET {'refresh_token': None, ...}
          Second test   SET {'refresh_token': 'test_refresh_token', ...}
          Second test   RUN test
```


I'm saying these are redundant

```
tap-salesforce config   {'refresh_token': None, ...}

           First test   SET {'refresh_token': None, ...}   # redundant
           First test   SET {'refresh_token': 'test_refresh_token', ...}
           First test   RUN test

          Second test   SET {'refresh_token': None, ...}   # redundant
          Second test   SET {'refresh_token': 'test_refresh_token', ...}
          Second test   RUN test

```

So we can simplify the test to just have

```
tap-salesforce config   {'refresh_token': None, ...}

           First test   SET {'refresh_token': 'test_refresh_token', ...}
           First test   RUN test

          Second test   SET {'refresh_token': 'test_refresh_token', ...}
          Second test   RUN test
```

tests/test_salesforce_lookback_window.py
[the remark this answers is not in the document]
kspeer825: I think Andy's point is that we could assert on just the first record. But for that assertion to be valid, we'd need to also assert that records are returned in order of ascending replication key value. Which I think is a better approach, but both are valid. 
Also it's always preferable to avoid assertTrue and assertFalse. They give less information than the comparison methods, even with the additional error msg.

luandy64 (changes requested):

tests/unittests/test_lookback_window.py
```diff
@@ -0,0 +1,197 @@
...
+    def login(self):
+        pass
+
+# mock args and return desired state, catalog and config file
+class MockParseArgs:
+    discover = None
+    properties = None
+    state = None
+    def __init__(self, config):
+        self.config = config
```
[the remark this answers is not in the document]
harshpatel4crest: Updated.

tests/unittests/test_lookback_window.py
```diff
@@ -0,0 +1,197 @@
...
+        self.assertEqual(start_date, '2021-01-02T00:00:00Z')
+
+    def test_no_lookback_window_subtraction_with_state__get_start_date(self):
+        """
+            Test case to verify state file date is not changed when we pass 'with_lookback=False'
+        """
+        # function call with apply lookback as 'False'
+        start_date = self.sf_without_lookback.get_start_date(self.mock_state, self.mock_catalog_entry, with_lookback=False)
+
+        # verify 20 seconds were subtracted from state file date
```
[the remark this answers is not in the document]
harshpatel4crest: Updated the comment.

tests/unittests/test_lookback_window.py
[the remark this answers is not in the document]
harshpatel4crest: Got it. Removed the code to reset `CONFIG` in the unittest.

tests/test_salesforce_lookback_window.py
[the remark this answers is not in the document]
harshpatel4crest: @kspeer825 @luandy64 Added assertion to verify we got records in ascending order and also updated the existing `is_between` assertion to verify that the 1st record is between the lookback date and the initial bookmark.

kspeer825 (approved):

tap_salesforce/salesforce/__init__.py
```diff
@@ -381,13 +387,25 @@ def _get_selected_properties(self, catalog_entry):
...
+        sync_start_date = singer.get_bookmark(state, catalog_entry['tap_stream_id'], replication_key) or self.default_start_date
+        # return bookmark from the state or start date if 'with_lookback' is False
+        if not with_lookback:
+            return sync_start_date
+
+        # if the state contains a bookmark, subtract the lookback window from the bookmark
+        if singer.get_bookmark(state, catalog_entry['tap_stream_id'], replication_key):
+            sync_start_date = singer_utils.strftime(singer_utils.strptime_with_tz(sync_start_date) - datetime.timedelta(seconds=self.lookback_window))
+
+        return sync_start_date
```
luandy64, line 408: Update this implementation to get the tests in `tests/unittests/test_lookback_window.py` passing

tap_salesforce/salesforce/__init__.py
```diff
@@ -381,13 +387,25 @@ def _get_selected_properties(self, catalog_entry):
                                             self.select_fields_by_default)]
 
 
-    def get_start_date(self, state, catalog_entry):
+    def get_start_date(self, state, catalog_entry, with_lookback=False):
+        """
+            if 'with_lookback' is False, then return start date or state file date
+            else subtract lookback window from the state file date and return
+                if no state file bookmark is found, do not subtract lookback window from the start date
+        """
```
luandy64, line 395: Update this comment to match the new implementation

tap_salesforce/salesforce/__init__.py
luandy64: One helpful thing might be helpful is that `self.lookback_window` is an integer or None, so we can use its value or fallback to zero

tap_salesforce/salesforce/__init__.py
```diff
@@ -235,6 +240,7 @@ def __init__(self,
         self.login_timer = None
         self.data_url = "{}/services/data/v52.0/{}"
         self.pk_chunking = False
+        self.lookback_window = lookback_window
 
         # validate start_date
         singer_utils.strptime(default_start_date)
```
luandy64, line 246: Update this to `singer_utils.strptime_to_utc` to get unit tests working again
- `singer_utils.strptime` is deprecated

The unit tests didn't catch this because we mocked away this function. My test removed the Mock and allowed this to fail properly.

---

The difference between [singer_utils.strptime](https://github.com/singer-io/singer-python/blob/f2b383324a3a97b98d6e29f69dbadae2632dbd2f/singer/utils.py#L58) and [singer_utils.strptime_to_utc](https://github.com/singer-io/singer-python/blob/f2b383324a3a97b98d6e29f69dbadae2632dbd2f/singer/utils.py#L61) is the parsing function used.

`strptime` needs a datetime format to function correctly - which results in the error that you see in Circle

This is also backwards compatible assuming we only write bookmarks with `singer-python` functions. This ensures that bookmarks are all converted to UTC first. And `singer_utils.strptime_to_utc` would just work because it reads a variety of date time formats (with and without fractional seconds) and then converts it to UTC.

---

Example to show that we can parse fractional seconds:
```python
>>> import dateutil.parser
>>> dateutil.parser.parse("2022-05-02T00:00:00.000000Z")
datetime.datetime(2022, 5, 2, 0, 0, tzinfo=tzlocal())
>>> dateutil.parser.parse("2022-05-02T00:00:00Z")
datetime.datetime(2022, 5, 2, 0, 0, tzinfo=tzlocal())
```

tap_salesforce/salesforce/bulk.py
```diff
@@ -107,7 +107,7 @@ def _can_pk_chunk_job(self, failure_message): # pylint: disable=no-self-use
 
     def _bulk_query(self, catalog_entry, state):
         job_id = self._create_job(catalog_entry)
-        start_date = self.sf.get_start_date(state, catalog_entry)
+        start_date = self.sf.get_start_date(state, catalog_entry, with_lookback=True)
```
luandy64, line 110: Remove `with_lookback` because it should not exist as a parameter in the new implementation

tap_salesforce/salesforce/rest.py
```diff
@@ -14,7 +14,7 @@ def __init__(self, sf):
         self.sf = sf
 
     def query(self, catalog_entry, state):
-        start_date = self.sf.get_start_date(state, catalog_entry)
+        start_date = self.sf.get_start_date(state, catalog_entry, with_lookback=True)
```
luandy64, line 17: Remove `with_lookback` because it should not exist as a parameter in the new implementation

README.md
```diff
@@ -40,6 +41,8 @@ The `start_date` is used by the tap as a bound on SOQL queries when searching fo
 
 The `api_type` is used to switch the behavior of the tap between using Salesforce's "REST" and "BULK" APIs. When new fields are discovered in Salesforce objects, the `select_fields_by_default` key describes whether or not the tap will select those fields by default.
 
+The `lookback_window` (in seconds) subtracts the desired amount of seconds from the bookmark to sync past data. Default value: 10 seconds.
```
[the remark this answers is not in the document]
luandy64: We should recommend a small value like `10` instead

luandy64 (changes requested):
# TDL-16115 Updated Requirements

Definition of Done:
1. Implement an optional `lookback_window` parameter
    - The tap should not use a lookback window unless it was configured to do so
2. Unit tests are passing
3. Integration tests are passing

tap_salesforce/salesforce/rest.py
harshpatel4crest: Removed `with_lookback` param from all places.

tap_salesforce/salesforce/bulk.py
harshpatel4crest: Removed `with_lookback` param from all places.

tap_salesforce/salesforce/__init__.py
harshpatel4crest: Thanks for the information. Updated `strptime` to `strptime_to_utc` and we will use **strptime_to_utc** from now onwards when needed.

tap_salesforce/salesforce/__init__.py
harshpatel4crest: Updated the comment as per the suggested implementation.

tap_salesforce/salesforce/__init__.py
harshpatel4crest: Thanks. Updated the code to only subtract lookback if the bookmark is present in the state and lookback is not `None`.

tap_salesforce/salesforce/__init__.py
```diff
@@ -381,13 +387,25 @@ def _get_selected_properties(self, catalog_entry):
                                             self.select_fields_by_default)]
 
 
-    def get_start_date(self, state, catalog_entry):
+    def get_start_date(self, state, catalog_entry, with_lookback=False):
```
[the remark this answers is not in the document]
harshpatel4crest: Removed `with_lookback` param from all places.

tap_salesforce/salesforce/__init__.py
```diff
@@ -210,7 +214,8 @@ def __init__(self,
                  is_sandbox=None,
                  select_fields_by_default=None,
                  default_start_date=None,
-                 api_type=None):
+                 api_type=None,
+                 lookback_window=DEFAULT_LOOKBACK_WINDOW):
```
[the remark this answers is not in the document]
harshpatel4crest: Added default value of `None` for **lookback_window** argument.

tap_salesforce/__init__.py
```diff
@@ -384,7 +384,8 @@ def main_impl():
             is_sandbox=CONFIG.get('is_sandbox'),
             select_fields_by_default=CONFIG.get('select_fields_by_default'),
             default_start_date=CONFIG.get('start_date'),
-            api_type=CONFIG.get('api_type'))
+            api_type=CONFIG.get('api_type'),
+            lookback_window=int(CONFIG.get('lookback_window', DEFAULT_LOOKBACK_WINDOW)))
```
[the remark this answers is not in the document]
harshpatel4crest: Got it. Updated lookback calculation as per the suggestion.

tap_salesforce/__init__.py
```diff
@@ -6,7 +6,7 @@
 from singer import metadata, metrics
 import tap_salesforce.salesforce
 from tap_salesforce.sync import (sync_stream, resume_syncing_bulk_query, get_stream_version)
-from tap_salesforce.salesforce import Salesforce
+from tap_salesforce.salesforce import Salesforce, DEFAULT_LOOKBACK_WINDOW
```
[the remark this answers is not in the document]
harshpatel4crest: Removed the import as it will not be required now.

harshpatel4crest:
Thanks, @luandy64 for a detailed review of the PRs. We got many things to learn from it and will apply them in the future.  We addressed the review comment for the remaining PR. Can you please re-review it?

**NOTE:** The build is failing as we need to add `lookback_window` to the report card for `platform.salesforce`.

luandy64 (commented):
The changes look good to me. I will approve when Circle is passing

luandy64 (approved):

PR #137, opened 2022-04-08
TDL-7301: Handle multiline `CRITICAL` error messages

harshpatel4crest:
# Description of change
TDL-7301: Handle multiline `CRITICAL` error messages
- Updated code to add `CRITICAL` for multiline error messages

# QA steps
 - [ ] automated tests passing
 
# Risks

# Rollback steps
 - revert this branch

### DIFF
--- a/tap_salesforce/__init__.py
+++ b/tap_salesforce/__init__.py
@@ -417,5 +417,6 @@ def main():
         LOGGER.critical(e)
         sys.exit(1)
     except Exception as e:
-        LOGGER.critical(e)
+        for error_line in str(e).splitlines():
+            LOGGER.critical(error_line)
         raise e
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -100,6 +100,12 @@ workflows:
             - tier-1-tap-user
           requires:
             - ensure_env
+      - run_unit_tests:
+          context:
+            - circleci-user
+            - tier-1-tap-user
+          requires:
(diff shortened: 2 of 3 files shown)


savan-chovatiya (approved):

tap_salesforce/__init__.py
```diff
@@ -417,5 +417,6 @@ def main():
         LOGGER.critical(e)
         sys.exit(1)
     except Exception as e:
-        LOGGER.critical(e)
```
dbshah1212, line 420: Will we get traceback after this change? In tap-facebook, they added Logger.exception specifically: https://github.com/singer-io/tap-facebook/blob/3318ad972a4c0df7b5f1c8bd57b942d5f6b39bf4/tap_facebook/__init__.py#L931

harshpatel4crest: Yes, we are getting traceback.

dbshah1212: As ask is to make it like a facebook, do we need to add Logger.exception specifically?

harshpatel4crest: I don't think so, the ask of this ticket was to Log `CRITICAL` for multiline error messages. And `LOGGER.exception` Logs `ERROR`.

dbshah1212 (approved):

luandy64 (approved):

RushT007 (approved):

PR #156, opened 2023-05-11
Add table reset testcase

bhuvana-talend:
# Description of change
(write a short description here or paste a link to JIRA)

# QA steps
 - [ ] automated tests passing
 
# Risks

# Rollback steps
 - revert this branch


tests/test_salesforce_full_table_reset.py
```diff
@@ -0,0 +1,89 @@
+import unittest
```
bhtowles, line 1: This appears to be a less rigorous FULL_TABLE test.  The concept of resetting a FULL_TABLE stream doesn't seem like a valid case since we do not expect to have a bookmark for a FULL_TABLE stream.  Since we do have FULL_TABLE streams for salesforce, id suggest replacing this test with a FULL_TABLE replication test.  There is an example of a SaaS full table test in the stripe test repo.

tests/test_salesforce_incremental_table_reset.py
```diff
@@ -0,0 +1,145 @@
...
+                # bookmarked states (actual values)
+                first_bookmark_value = first_bookmark_key_value.get(replication_key)
+                second_bookmark_value = second_bookmark_key_value.get(replication_key)
+
+                # bookmarked values as utc for comparing against records
+                first_bookmark_value_utc = self.convert_state_to_utc(first_bookmark_value)
+                second_bookmark_value_utc = self.convert_state_to_utc(second_bookmark_value)
+
+                # Verify the second sync bookmark is Equal to the first sync bookmark
+                self.assertEqual(second_bookmark_value, first_bookmark_value) # assumes no changes to data during test
```
bhtowles, line 116: I'd suggest either using our filter pattern here or asserting greaterEqual to allow for data coming in between syncs as we have seen this be a problem in the past.

tests/test_salesforce_incremental_table_reset.py
```diff
@@ -0,0 +1,145 @@
...
+                for record in second_sync_messages:
+                    replication_key_value = record.get(replication_key)
+                    self.assertLessEqual(replication_key_value, second_bookmark_value_utc,
+                                         msg="Second sync bookmark was set incorrectly, a record with a greater rep key value was synced")
+
+                # Verify the number of records in the 2nd sync is equal to first for the stream that is reset and is less then the firsa for the restt
+                if stream != 'User':
+                    self.assertLess(second_sync_count, first_sync_count)
+                else:
+                    self.assertEqual(second_sync_count, first_sync_count)
```
bhtowles, line 142: Same issue here with new data.

tests/test_salesforce_incremental_table_reset.py
```diff
@@ -0,0 +1,145 @@
...
+        return datetime.datetime.strftime(date_object_utc, "%Y-%m-%dT%H:%M:%SZ")
+
+    def calculated_states_by_stream(self, current_state):
+        """
+        Returns the streams from the current state
+        """
+
+        stream_to_current_state = {stream : bookmark.get(self.expected_replication_keys()[stream].pop())
+                                   for stream, bookmark in current_state['bookmarks'].items()}
+        return stream_to_current_state
```
bhtowles, line 42: I don't think I'm following this test method.  The name implies a calculation is being done?

tests/test_salesforce_full_table_reset.py
bhuvana-talend: Ok, Will create a separate card for full table replication and implement it. Will remove the full_table_reset from this PR.

tests/test_salesforce_incremental_table_reset.py
bhuvana-talend: Changed the assertion to greaterEqual

tests/test_salesforce_incremental_table_reset.py
bhuvana-talend: For User stream, Changed the assertion to greaterEqual

tests/test_salesforce_incremental_table_reset.py
bhuvana-talend: Changed the method name to "get_states_by_stream" to be reflective of what the method actually does.

tests/test_salesforce_incremental_table_reset.py
```diff
@@ -0,0 +1,147 @@
...
+                                         msg="First sync bookmark was set incorrectly, a record with a greater rep key value was synced")
+
+                # Verify the second sync bookmark value is the max replication key value for a given stream
+                for record in second_sync_messages:
+                    replication_key_value = record.get(replication_key)
+                    self.assertLessEqual(replication_key_value, second_bookmark_value_utc,
+                                         msg="Second sync bookmark was set incorrectly, a record with a greater rep key value was synced")
+
+                # Verify the number of records in the 2nd sync is greater or equal to first 
+                # If no new records in between 2 sync, the second will be eual to first for 'User' and will be less for the other streams
```
bhtowles, line 140: typo

bhtowles (changes requested):

bhtowles:
I may have been incorrect about full table not supporting interrupts for SaaS taps.  In order to know if this is supported or not we need to inspect the code or run the test and inspect state during a sync for a full table.  If we save last pk fetched or other bookmark info for full table streams that are currently syncing then it is possible to resume a full table interrupted sync and we can add the test you have above.  If we do not save any bookmark state for currently syncing full table streams then there is nothing to test.

tests/test_salesforce_full_table_reset.py
bhuvana-talend: Removed the file

tests/test_salesforce_incremental_table_reset.py
bhuvana-talend: Removed the method and making the format change in the main method

tests/test_salesforce_incremental_table_reset.py
```diff
@@ -0,0 +1,83 @@
+import unittest
```
bhtowles, line 1: There seems to be a lot of changes here.  The test went from 145 lines in commit f1689d398be8d0e704f1fd41fd5c2645c7cdd621. To Only 83 lines here.  There are at least 8 assertions in the previous test.  This one has only 2.

tests/test_salesforce_incremental_table_reset.py
```diff
@@ -0,0 +1,83 @@
...
+        conn_id = connections.ensure_connection(self)
+
+        # Run in check mode
+        found_catalogs = self.run_and_verify_check_mode(conn_id)
+
+        # Select only the expected streams tables
+        expected_streams = self.expected_sync_streams()
+        catalog_entries = [ce for ce in found_catalogs if ce['tap_stream_id'] in expected_streams]
+        self.select_all_streams_and_fields(conn_id, catalog_entries)
+        streams_replication_methods = {stream: self.FULL_TABLE
```
bhtowles, line 49: FULL table vs INCREMENTAL?

tests/test_salesforce_incremental_table_reset.py
```diff
@@ -0,0 +1,83 @@
...
+            # Remove stream User to reset
+            if stream != 'User':
+                new_states['bookmarks'][stream] = {replication_key: stream_to_current_state[stream]}
+        menagerie.set_state(conn_id, new_states)
+
+        # SYNC 2
+        second_sync_record_count = self.run_and_verify_sync(conn_id)
+        second_sync_bookmarks = menagerie.get_state(conn_id)
+
+        #Verify if the 2 syncs returned the same set of records
```
bhtowles, line 72: Comment does not match the assertion.  One mentions a record comparison while the other does a bookmark comparison.  Also the bookmark comparison is unfiltered so it can fail if new data comes in between syncs.

tests/test_salesforce_incremental_table_reset.py
```diff
@@ -0,0 +1,83 @@
...
+        self.assertEqual(first_sync_bookmarks, second_sync_bookmarks)
+
+        # Test by stream
+        for stream in expected_streams:
+            with self.subTest(stream=stream):
+                # record counts
+                first_sync_count = first_sync_record_count.get(stream, 0)
+                second_sync_count = second_sync_record_count.get(stream, 0)
+
+                # Verify the number of records in the 2nd sync is the same as the first
```
bhtowles, line 82: Counts are also unfiltered and can vary if new data comes between syncs.  Also what happened to the other 6 assertions?

bhtowles (changes requested):

tests/test_salesforce_incremental_table_reset.py
bhuvana-talend: Added filtering to take care of record inserts between 2 syncs

tests/test_salesforce_incremental_table_reset.py
bhuvana-talend: Added filtering to take care of record inserts between 2 syncs

tests/test_salesforce_incremental_table_reset.py
```diff
@@ -0,0 +1,83 @@
+import unittest
+import datetime
+import pytz
+
+from tap_tester import runner, menagerie, connections
+
+from base import SalesforceBaseTest
+
+
+class SalesforceFullTableReset(SalesforceBaseTest):
```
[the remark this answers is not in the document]
bhuvana-talend: Changed the names to match

tests/test_salesforce_incremental_table_reset.py
bhuvana-talend: Added back the needed assertions

tests/test_salesforce_incremental_table_reset.py
bhuvana-talend: Added filtering to take care of record inserts between 2 syncs

tests/test_salesforce_incremental_table_reset.py
```diff
@@ -0,0 +1,141 @@
...
+
+        # Run a sync job using orchestrator
+        first_sync_record_count = self.run_and_verify_sync(conn_id)
+        first_sync_records = runner.get_records_from_target_output()
+
+        first_sync_bookmarks = menagerie.get_state(conn_id)
+
+        # UPDATE STATE for Table Reset
+        new_states = {'bookmarks': dict()}
+        stream_to_current_state = {stream : bookmark.get(self.expected_replication_keys()[stream].pop())
```
bhtowles, line 84: This seems like a lot of lines of code to remove one bookmark.  I've seen deep copy used in other tests so you may want to look into why that may be preferred.  What about something like this?
```suggestion
        new_states =  first_sync_bookmarks
        new_states['bookmarks'].pop('Users')
        menagerie.set_state(conn_id, new_states)
```

bhtowles: Also, I think it would be more maintainable and readable if we set store the reset stream in a variable and set it.  Then everywhere where we are referencing the 'Users' stream we can instead reference the name of the variable.  Maybe call it`reset_stream`.  Then it will be obvious why this stream is being treated differently and if we ever want to change the stream we only need to change it in one place.

tests/test_salesforce_incremental_table_reset.py
```diff
@@ -0,0 +1,141 @@
...
+
+                # bookmarked states (top level objects)
+                first_bookmark_key_value = first_sync_bookmarks.get('bookmarks').get(stream)
+                second_bookmark_key_value = second_sync_bookmarks.get('bookmarks').get(stream)
+
+                # bookmarked states (actual values)
+                first_bookmark_value = first_bookmark_key_value.get(replication_key)
+                second_bookmark_value = second_bookmark_key_value.get(replication_key)
+
+                filtered_synced_records = [
```
bhtowles, line 126: I'd like to include which sync is being filtered in the variable name for readability.  Maybe filtered_second_sync_records?

bhtowles (changes requested):

tests/test_salesforce_incremental_table_reset.py
bhuvana-talend: changed the variable name to filtered_second_sync_records for better readability

tests/test_salesforce_incremental_table_reset.py
bhuvana-talend: also made a deep copy to have a clean code

bhtowles (approved):
Nice!

PR #159, opened 2023-06-08
Implement all_fields test using new framework-tdl-7499

bhuvana-talend:
# Description of change
https://jira.talendforge.org/browse/TDL-7499
Implement all_fields_test using the new tap-tester framework.
Tested with streams that also has custom fields 
Also modified tap-tester/base_suite_tests/base_case.py : get_all_streams_and_fields to exclude unsupported fields

# QA steps
 - [ ] automated tests passing
 
# Risks

# Rollback steps
 - revert this branch


tests/sfbase.py
```diff
@@ -0,0 +1,937 @@
+"""
```
bhtowles, line 1: I am not up to speed on these split taps with two base files yet.  Will save this one for Monday unless you and Jyothi can get me up to speed.

bhuvana-talend: Sure, basically copied over base.py and 
instead of 
from tap_tester.base_case import BaseCase
have this: from tap_tester.base_suite_tests.base_case import BaseCase
- Removed unnecessary methods 
- commented out some fields in metadata that is not discovered by discovery to make the assertion pass for discovery

bhtowles: So is the plan to maintain two base files until the full suite is cut over then remove / rename one?  Why not just extend / modify one base file?

bhuvana-talend: yeah, the idea is to have 2 files still fuly cutover.
Not sure if it is possible to use the same file as it can create conflicts since there are some methods in both tap_tester.base_case and also tap_tester.base_suite_tests.base_case

tests/test_salesforce_custom_fields_test.py
```diff
@@ -0,0 +1,43 @@
...
+    @staticmethod
+    def name():
+        return "tt_sf_custom_fields_test"
+
+    def streams_to_test(self):
+        streams = {'Account', 'Contact'}
+        return streams
+
+    def test_custom_fields( self ):
+        expected_streams = self.selected_fields
```
bhtowles, line 28: It does not appear that we are doing any filtering up front on field selection so even though this is the custom fields test I believe we will be selecting, and replicating all fields.

Selecting all fields but then only testing that we have selected and replicated the custom ones may result in a similar test but will be much less efficient than just selecting, replicating, and verifying the portion of streams and fields we want.  It could also possibly mask a field selection bug.

tests/test_salesforce_custom_fields_test.py
```diff
@@ -0,0 +1,43 @@
...
+        expected_fields = set()
+        for stream in self.streams_to_test():
+            for value in expected_streams.get(stream, set()):
+                if value.endswith("__c"):
+                    expected_fields.add(value)
```
bhtowles, line 33: Since we are building a set of fields should we implement a counter to check for any duplicates and assert the count == length of the set?

Also I'd like to use more descriptive names.  Maybe something like:
selected_custom_fields
replicated_custom_fields
or 
expected_custom_fields
actual_custom_fields

tests/test_salesforce_non_custom_fields_test.py
```diff
@@ -0,0 +1,43 @@
+"""
+Test that with no fields selected for a stream automatic fields and non custom fields  are still replicated
```
bhtowles, line 2: Doc strings for both tests need to be updated.

Same comments as custom test.

bhtowles (changes requested):

tests/test_salesforce_custom_fields_test.py
```diff
@@ -0,0 +1,43 @@
...
+
+        actual = set()
+        for stream in self.streams_to_test():
```
bhtowles, line 36: We also discussed an assertion that if the set of expected or actual fields is empty then the test should fail.  So we were going to assert each set is True.  This makes sure that we don't compare two empty sets and pass when there are no custom fields.

tests/test_salesforce_all_fields.py
```diff
@@ -0,0 +1,26 @@
+"""
```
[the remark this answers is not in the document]
bhuvana-talend: Sure, I'll remove this file

tests/test_salesforce_custom_fields_test.py
bhuvana-talend: Renamed the fields with more descriptive names

tests/test_salesforce_custom_fields_test.py
bhuvana-talend: I have overridden the methid streams_to_selected_fields to return a dict of stream and custom_fields for custom fields testcase and a dict of stream and non_custom_fields for non custom fields testcase

tests/test_salesforce_custom_fields_test.py
```diff
@@ -0,0 +1,43 @@
+"""
```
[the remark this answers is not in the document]
bhuvana-talend: Done

bhtowles:
Re-started commit work flow from start as last test failed due to quota to see if tests will pass.

tests/test_salesforce_all_fields_non_custom.py
```diff
@@ -0,0 +1,56 @@
...
+    @staticmethod
+    def streams_to_selected_fields():
+        return SFBaseTest.non_custom_fields
+
+    def test_non_custom_fields( self ):
+        for stream in self.streams_to_selected_fields():
+            expected_non_custom_fields = self.streams_to_selected_fields().get(stream, set() )
+            replicated_non_custom_fields = self.actual_fields.get(stream, set() )
+
+            self.assertIsNotNone( replicated_non_custom_fields, msg = f"Replication didn't return any non-custom fields for stream {stream}" )
```
bhtowles, line 36: Maybe while we are switching from hard coded to programmatically building the list of fields we can test for both custom and non custom field selection in both tests like this for completeness.  For now we can make this a TODO?
So in the non_custom test we have:
  self.assertIsNotNone(replicated_non_custom_fields)
  self.assertIsNone(replicated_custom_fields.difference(automatic_fields))
And in the custom test we have:
  self.assertIsNone(replicated_non_custom_fields)
  self.assertIsNotNone(replicated_custom_fields)

tests/test_salesforce_all_fields_non_custom.py
```diff
@@ -0,0 +1,56 @@
...
+
+            self.assertSetEqual(expected_non_custom_fields, replicated_non_custom_fields,
+                 logging=f"verify all non custom fields are replicated for stream {stream}")
+
+    #Override this method to exclude automatic fields in the assertion
```
bhtowles, line 41: Yes, I think this explains why we are over-riding the previous test.  We don't need to over-ride this one because the automatic fields are not custom fields so they are included in expectations.  But in the custom fields test they are excluded from selection.  So if we union the expectation in the base suite test to always include automatic fields then we should not have to over-ride either of these tests (I hope 😄)

bhtowles (changes requested):

tests/test_salesforce_all_fields_non_custom.py
```diff
@@ -0,0 +1,50 @@
...
+        return "tt_sf_all_fields_non_custom"
+
+    def streams_to_test(self):
+        streams = {'Account', 'Contact'}
+        return streams
+
+    def test_non_custom_fields( self ):
+        for stream in self.streams_to_selected_fields():
+            expected_non_custom_fields = self.streams_to_selected_fields().get(stream, set() )
+            replicated_non_custom_fields = self.actual_fields.get(stream, set() )
```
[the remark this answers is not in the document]
bhuvana-talend: modifed field names

tests/test_salesforce_all_fields_custom.py
```diff
@@ -0,0 +1,59 @@
...
+                # gather expectations
+                expected_all_keys = self.selected_fields.get(stream, set())
+
+                # gather results
+                fields_replicated = self.actual_fields.get(stream, set())
+                automatic_fields = self.expected_automatic_fields ( stream )
+
+                # verify that all fields are sent to the target
+                # test the combination of all records
+                self.assertSetEqual(fields_replicated.difference(automatic_fields), expected_all_keys,
```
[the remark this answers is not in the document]
bhuvana-talend: Modified tap-tester/base_suite_tests/all_fields_test.py to include the automatic fields in expected_keys

tests/test_salesforce_all_fields_non_custom.py
bhuvana-talend: Added notnone assertion and made a note in the other card for the other assertion

tests/test_salesforce_all_fields_non_custom.py
bhuvana-talend: Got it, made the change and removed the unnecessary assertion

tests/test_salesforce_all_fields_custom.py
```diff
@@ -0,0 +1,59 @@
...
+    @staticmethod
+    def streams_to_selected_fields():
+        return SFBaseTest.custom_fields
+
+    def test_custom_fields( self ):
+        for stream in self.streams_to_selected_fields():
+            expected_custom_fields = self.streams_to_selected_fields().get(stream, set() )
+            actual_custom_fields = self.actual_fields.get(stream, set() )
+
+            self.assertIsNotNone( actual_custom_fields, msg = f"Replication didn't return any custom fields for stream {stream}" )
```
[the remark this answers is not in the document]
bhuvana-talend: removed the fields verification

tests/test_salesforce_all_fields_non_custom.py
```diff
@@ -0,0 +1,47 @@
...
+        return streams
+
+    @staticmethod
+    def streams_to_selected_fields():
+        return SFBaseTest.non_custom_fields
+
+    def test_non_custom_fields( self ):
+        for stream in self.streams_to_selected_fields():
+            expected_non_custom_fields = self.streams_to_selected_fields().get(stream, set() )
+            replicated_non_custom_fields = self.actual_fields.get(stream, set() ).difference(self.expected_automatic_fields(stream))
```
bhtowles, line 34: We don't want to subtract automatic fields here because the automatic fields are non custom fields.

tests/test_salesforce_all_fields_non_custom.py
```diff
@@ -0,0 +1,47 @@
...
+
+            #Verify at least one non-custom field is replicated
+            self.assertIsNotNone( replicated_non_custom_fields, msg = f"Replication didn't return any non-custom fields for stream {stream}" )
+
+            #Verify ustom fields are not replicated by checking the field name
+            self.assertFalse( self.verify_custom_fields( replicated_non_custom_fields ), "Replicated some fields that are custom fields for stream {stream}" )
```
bhtowles, line 40: This will pass as long as one field in the collection is not custom.  From the sounds of the comment I think you want to verify that non of the fields are custom.  I think you may need to update the verify_custom_fields logic if you only want to use it like this.  Or have two separate functions, one to verify all fields are custom and one to verify all fields are not custom.  Also, extra space around the parenthesis.

bhtowles (changes requested):

tests/test_salesforce_all_fields_non_custom.py
bhuvana-talend: The reason to subtract automatic fields - if the replicated fields contain only automatic fields, my assertion for Not none in the next line will not verify if fields other than automatic fields are replicated or not.

tests/test_salesforce_all_fields_non_custom.py
bhuvana-talend: Modified the method to return the count of custom and non-custom fields and make assertion on the counts instead. Also changed the name of the method and moved it to sfbase.py

tests/test_salesforce_all_fields_non_custom.py
bhtowles: The automatic fields are non custom fields.  The test is currently making a sub-set of non_custom fields.  Do we need to track both groups?  All non custom fields and then all non custom fields that aren't automatic?  I'm not sure we need to do this.  We could have two assertions if you want to check both cases:
self.assertIsNotNone(replicated_non_custom_fields, msg)
self.assertIsNotNone(replicated_non_custom_fields.difference(self.expected_automatic_fields(stream)), msg)

tests/sfbase.py
```diff
@@ -26,6 +34,17 @@ def get_type():
...
+    @staticmethod
+    def count_custom_non_custom_fields(fields):
+        custom = 0
+        non_custom =0
+        for field in fields:
+            if not field.endswith("__c"):
+                non_custom += 1
+            else:
+                custom += 1
+        return ( custom, non_custom )
```
bhtowles, line 46:
```suggestion
        return (custom, non_custom)
```

tests/test_salesforce_all_fields_custom.py
```diff
@@ -0,0 +1,48 @@
...
+        streams = {'Account', 'Contact'}
+        return streams
+
+    @staticmethod
+    def streams_to_selected_fields():
+        return SFBaseTest.custom_fields
+
+    def test_custom_fields( self ):
+        for stream in self.streams_to_selected_fields():
+            expected_custom_fields = self.streams_to_selected_fields().get(stream, set() )
```
bhtowles, line 34:
```suggestion
            expected_custom_fields = self.streams_to_selected_fields().get(stream, set())
```

tests/test_salesforce_all_fields_custom.py
```diff
@@ -0,0 +1,48 @@
...
+            replicated_custom_fields = self.actual_fields.get(stream, set() ).difference(self.expected_automatic_fields(stream))
+
+            #Verify at least one custom field is replicated
+            self.assertIsNotNone( replicated_custom_fields, msg = f"Replication didn't return any custom fields for stream {stream}" )
```
bhtowles, line 38:
```suggestion
            self.assertIsNotNone(replicated_custom_fields, msg = f"Replication didn't return any custom fields for stream {stream}")
```

tests/test_salesforce_all_fields_custom.py
```diff
@@ -0,0 +1,48 @@
...
+
+            #Verify only custom fields are replicated by checking the field name
+            num_custom, num_non_custom = self.count_custom_non_custom_fields(replicated_custom_fields)
+            self.assertEqual( num_non_custom, 0, "Replicated some fields that are not custom fields for stream {stream}" )
```
bhtowles, line 42:
```suggestion
            self.assertEqual(num_non_custom, 0, 
              "Replicated some fields that are not custom fields for stream {stream}")
```

tests/test_salesforce_all_fields_non_custom.py
```diff
@@ -0,0 +1,48 @@
...
+
+    def streams_to_test(self):
+        streams = {'Account', 'Contact'}
+        return streams
+
+    @staticmethod
+    def streams_to_selected_fields():
+        return SFBaseTest.non_custom_fields
+
+    def test_non_custom_fields( self ):
```
bhtowles, line 31:
```suggestion
    def test_non_custom_fields(self):
```

tests/test_salesforce_all_fields_non_custom.py
```diff
@@ -0,0 +1,48 @@
...
+        for stream in self.streams_to_selected_fields():
+            expected_non_custom_fields = self.streams_to_selected_fields().get(stream, set() )
```
bhtowles, line 33:
```suggestion
            expected_non_custom_fields = self.streams_to_selected_fields().get(stream, set())
```

tests/test_salesforce_all_fields_non_custom.py
```diff
@@ -0,0 +1,48 @@
...
+            replicated_non_custom_fields = self.actual_fields.get(stream, set() ).difference(self.expected_automatic_fields(stream))
+
+            #Verify at least one non-custom field is replicated
+            self.assertIsNotNone( replicated_non_custom_fields, msg = f"Replication didn't return any non-custom fields for stream {stream}" )
```
bhtowles, line 37:
```suggestion
            self.assertIsNotNone(replicated_non_custom_fields, msg = f"Replication didn't return any non-custom fields for stream {stream}")
```

bhtowles, line 37:
```suggestion
            self.assertIsNotNone(replicated_non_custom_fields, msg = f"Replication didn't return any non-custom fields for stream {stream}")
```

tests/test_salesforce_all_fields_non_custom.py
```diff
@@ -0,0 +1,48 @@
...
+
+            #Verify ustom fields are not replicated by checking the field name
+            num_custom, num_non_custom = self.count_custom_non_custom_fields(replicated_non_custom_fields)
+            self.assertEqual(num_custom, 0, "Replicated some fields that are custom fields for stream {stream}" )
```
bhtowles, line 41:
```suggestion
            self.assertEqual(num_custom, 0, "Replicated some fields that are custom fields for stream {stream}")
```

bhtowles (changes requested):

tests/test_salesforce_all_fields_non_custom.py
bhtowles: I think this is the test that is still failing.  I suggest not removing automatic fields from replicated_non_custom_fields.  Then we can add two more assertions 

1. The first assertion to make sure that the count of non custom fields is greater than the count of automatic fields to make sure that we are replicating more than just the automatic fields.
2. To make sure the number of non custom fields is equal to the number of replicated fields to make sure we didn't miss any.

tests/sfbase.py
```diff
@@ -900,3 +918,241 @@ def setUpClass(cls):
...
+
+    def get_non_custom_fields(self, found_catalogs, conn_id):
+        """ List all the non_custom_fields for each stream"""
+        non_custom_fields = {}
+        for stream in self.streams_to_test():
+            with self.subTest(stream=stream):
+                catalog = [catalog for catalog in found_catalogs if catalog["stream_name"] == stream][0]
+                schema = menagerie.get_annotated_schema(conn_id, catalog['stream_id'])["annotated-schema"]
+                non_custom_fields[stream] = {key for key in schema['properties'].keys() if not key.endswith("__c")
+                                             and  schema['properties'][key]['inclusion']!="unsupported"}
```
luandy64, line 941: 1. There's two spaces between `and` and `schema`
   - There should be one 
2. There's no spaces around `!=` 
   - There should be one on each side

Optional: I personally think if there's an `if` in a comprehension, it should get its own line and every `and` after that too. To me, it helps separate the filtering from the looping
```python
non_custom_fields[stream] = {key 
                             for key in schema['properties'].keys() 
                             if not key.endswith("__c")
                             and schema['properties'][key]['inclusion'] != "unsupported"}
```

tests/sfbase.py
```diff
@@ -900,3 +918,241 @@ def setUpClass(cls):
...
+    @staticmethod
+    def count_custom_non_custom_fields(fields):
+        custom = 0
+        non_custom =0
+        for field in fields:
+            if not field.endswith("__c"):
+                non_custom += 1
+            else:
+                custom += 1
+        return (custom, non_custom)
```
luandy64, line 953: This is a duplicate of lines 40-49

tests/sfbase.py
```diff
@@ -900,3 +918,241 @@ def setUpClass(cls):
...
+             'ActiveProfileMetric','Calendar',
+             'ContentWorkspacePermission',
+             'CampaignMemberStatus',
+             'Community',
+             'Contact',
+             'Case',
+             'ClientBrowser',
+             'BusinessHours',
+             'ContentWorkspace',
+             'Campaign','FieldPermissions',
```
luandy64, line 979: Any reason these are on the same line?

tests/sfbase.py
```diff
@@ -900,3 +918,241 @@ def setUpClass(cls):
...
+        returns True if stream is unsupported by BULK API
+
+        BULK API does not support any streams that are unsupported by the REST API and
+        in addition does not support the streams listed below.
+        """
+        return stream in self.get_unsupported_by_bulk_api()
+
+    def partition_streams(self,list_of_streams):
+
+        weekday = dt.weekday(dt.now())  # weekdays 0-6, Mon-Sun
```
luandy64, line 1137: Another way to write this is
```python
dt.now().weekday()
```

tests/test_salesforce_all_fields_custom.py
```diff
@@ -0,0 +1,49 @@
+"""
+Test that with only custom fields selected for a stream automatic fields and custom fields  are still replicated
+"""
+import copy
```
luandy64, line 4: This is unused

tests/test_salesforce_all_fields_custom.py
```diff
@@ -0,0 +1,49 @@
...
+from tap_tester.base_suite_tests.all_fields_test import AllFieldsTest
+from sfbase import SFBaseTest
+from tap_tester.logger import LOGGER
```
luandy64, line 7: This is unused

tests/test_salesforce_all_fields_custom.py
```diff
@@ -0,0 +1,49 @@
...
+
+            #Verify at least one custom field is replicated
+            self.assertIsNotNone(replicated_custom_fields.difference(self.expected_automatic_fields(stream)),
+                                 msg = f"Replication didn't return any custom fields for stream {stream}")
+
+            #Verify that custom and automatic fields are replicated
+            self.assertSetEqual(expected_custom_fields, replicated_custom_fields,
+                                logging=f"verify all fields are replicated for stream {stream}")
+
+            #Verify that only custome fields are replicated besides automatic fields
```
luandy64, line 46:
```suggestion
            #Verify that only custom fields are replicated besides automatic fields
```

tests/test_salesforce_all_fields_custom.py
```diff
@@ -0,0 +1,49 @@
...
+            num_custom, num_non_custom = self.count_custom_non_custom_fields(replicated_custom_fields)
```
luandy64, line 47: Typically in python, when a value is not used, we assign it to `_`
```suggestion
            _, num_non_custom = self.count_custom_non_custom_fields(replicated_custom_fields)
```

tests/test_salesforce_all_fields_custom_rest.py
```diff
@@ -0,0 +1,48 @@
+"""
+Test that with only custom fields selected for a stream automatic fields and custom fields  are still replicated
+"""
+import copy
+from tap_tester.base_suite_tests.all_fields_test import AllFieldsTest
+from sfbase import SFBaseTest
+from tap_tester.logger import LOGGER
+
+
+class SFCustomFieldsTest(AllFieldsTest, SFBaseTest):
```
luandy64, line 10: Is this allowed to be the same class name as the another test?

`test_salesforce_all_fields_custom.py` defines the same class

tests/test_salesforce_all_fields_custom_rest.py
```diff
@@ -0,0 +1,48 @@
...
+    def streams_to_selected_fields(self):
+        found_catalogs = AllFieldsTest.found_catalogs
+        conn_id = AllFieldsTest.conn_id
+        custom_fields = self.get_custom_fields(found_catalogs, conn_id)
+        return custom_fields
+
+    def test_all_fields_for_streams_are_replicated(self):
+        for stream in self.streams_to_test():
+
+            expected_custom_fields = self.selected_fields.get(stream, set()).union(self.expected_automatic_fields(stream))
```
luandy64, line 34: This line was wrapped in the `test_salesforce_all_fields_custom.py` file, should it be wrapped here?

tests/test_salesforce_all_fields_custom_rest.py
```diff
@@ -0,0 +1,48 @@
...
+            replicated_custom_fields = self.actual_fields.get(stream, set())
+
+            #Verify at least one custom field is replicate
```
luandy64, line 37:
```suggestion
            #Verify at least one custom field is replicated
```
this makes it match the other file

tests/test_salesforce_all_fields_custom_rest.py
```diff
@@ -0,0 +1,48 @@
...
+            self.assertIsNotNone(replicated_custom_fields.difference(self.expected_automatic_fields(stream)),
```
luandy64, line 38: Same issue here

tests/test_salesforce_all_fields_custom_rest.py
```diff
@@ -0,0 +1,48 @@
...
+                                 msg = f"Replication didn't return any custom fields for stream {stream}")
+
+            #Verify that custom and automatic fields are replicated
+            self.assertSetEqual(expected_custom_fields, replicated_custom_fields,
+                                logging=f"verify all fields are replicated for stream {stream}")
+
+            #Verify that only custome fields are replicated besides automatic fields
+            num_custom, num_non_custom = self.count_custom_non_custom_fields(replicated_custom_fields)
```
luandy64, line 46: Same comment here

tests/test_salesforce_all_fields_non_custom.py
```diff
@@ -0,0 +1,48 @@
...
+    salesforce_api = 'BULK'
+
+    @staticmethod
+    def name():
+        return "tt_sf_all_fields_non_custom"
+
+    def streams_to_test(self):
+        if self.partitioned_streams:
+            return self.partitioned_streams
+        return self.partition_streams(self.get_streams_with_data())
```
luandy64, line 20: This function is repeated 4 times and should be moved to `sfbase`

tests/test_salesforce_all_fields_non_custom.py
```diff
@@ -0,0 +1,48 @@
...
+            #Verify at least one non-custom field is replicated
+            self.assertIsNotNone(replicated_non_custom_fields,
+                                 msg = f"Replication didn't return any non-custom fields for stream {stream}")
+
+            #verify that all the non_custom fields are replicated
+            self.assertEqual(replicated_non_custom_fields, expected_non_custom_fields,
+                             msg = f"All non_custom fields are not no replicated for stream {stream}")
+
+            #verify that automatic fields are also replicated along with non_custom_fields
+            self.assertTrue(self.expected_automatic_fields(stream).issubset(replicated_non_custom_fields,
```
luandy64, line 42:
```suggestion
            self.assertTrue(self.expected_automatic_fields(stream).issubset(replicated_non_custom_fields),
```
This missing paren is causing CI to fail

tests/test_salesforce_sync_canary.py
```diff
@@ -46,30 +44,16 @@ def test_run(self):
...
+        allowed_catalogs = [catalog for catalog in found_catalogs
+                            if catalog['stream_name'] in expected_streams]
 
         self.select_all_streams_and_fields(conn_id, allowed_catalogs)
-
         # Run sync
         menagerie.set_state(conn_id, {})
-        _ = self.run_and_verify_sync(conn_id)
+        record_count_by_stream = self.run_and_verify_sync_mode(conn_id)
+        actual_streams_with_data ={stream for stream in record_count_by_stream.keys()
```
luandy64, line 56: Just so you know, looping over a dictionary gives you the `.keys()` by default

tests/test_salesforce_all_fields_non_custom_rest.py
```diff
@@ -0,0 +1,46 @@
...
+        return self.partition_streams(self.get_streams_with_data())
+
+    def streams_to_selected_fields(self):
+        found_catalogs = AllFieldsTest.found_catalogs
+        conn_id = AllFieldsTest.conn_id
+        non_custom_fields = self.get_non_custom_fields(found_catalogs, conn_id)
+        return non_custom_fields
+
+    def test_non_custom_fields(self):
+        for stream in self.streams_to_selected_fields():
```
luandy64, line 29: Should this be wrapped in a `subTest` so that in one run you can see all of the failing streams?

tests/sfbase.py
```diff
@@ -900,3 +918,241 @@ def setUpClass(cls):
 
         if missing_envs:
             raise Exception("set environment variables")
+
+    def get_custom_fields(self, found_catalogs, conn_id):
+        """ List all the custom_fields for each stream"""
+        custom_fields = {}
+        for stream in self.streams_to_test():
+            with self.subTest(stream=stream):
```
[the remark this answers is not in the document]
JYOTHINARAYANSETTY: Sure, will do that

tests/sfbase.py
JYOTHINARAYANSETTY: Will remove

tests/sfbase.py
JYOTHINARAYANSETTY: that was a miss , will fix it

tests/test_salesforce_all_fields_non_custom.py
JYOTHINARAYANSETTY: All these tests are implement in the new tap-tester framework and the All_fields test needs this method to be present in the test as it is a abstract method in tap-tester all_fields test , if I understood it right. Do let me know if thats not the case.
https://github.com/stitchdata/tap-tester/blob/b280a171ce2ecc565d54f96bbef7cac9fe4a61ec/tap_tester/base_suite_tests/all_fields_test.py#L49

luandy64: I think if `SFBaseTest` implements it then `SFNonCustomFieldsTest` doesn't have to

luandy64: Let me try to write up a quick test

bhtowles (approved):

PR #160, opened 2023-06-12
Update discovery streams, June 2023

bhtowles:
# Description of change
Update discovery streams, June 2023

# QA steps
 - [ ] automated tests passing
 
# Risks

# Rollback steps
 - revert this branch

### DIFF
--- a/tests/base.py
+++ b/tests/base.py
@@ -490,7 +490,7 @@ class SalesforceBaseTest(BaseCase):
             'LoginHistory': {self.PRIMARY_KEYS: {'Id'}, self.REPLICATION_KEYS: {'LoginTime'},self.REPLICATION_METHOD: self.INCREMENTAL,},
             'LoginIp': incremental_created_date,
             'LogoutEvent': default_full,  # new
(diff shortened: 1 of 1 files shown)


tests/base.py
```diff
@@ -490,7 +490,7 @@ def expected_metadata(self):
             'LoginHistory': {self.PRIMARY_KEYS: {'Id'}, self.REPLICATION_KEYS: {'LoginTime'},self.REPLICATION_METHOD: self.INCREMENTAL,},
             'LoginIp': incremental_created_date,
             'LogoutEvent': default_full,  # new
-            'MLField': default,  # removed # 6/13/2022 added back 7/10/2022
+            # 'MLField': default,  # removed 6/13/2022 added 7/10/2022, removed 06/12/2023
```
bhuvana-talend, line 493: Looks like this stream is added and removed couple of times in the past. Don't understand why this can happen. Is it only in test environment ?

bhtowles: No this is the response from the API.  Production should get the same results as our test environment.

bhuvana-talend (approved):

PR #161, opened 2023-06-13
Implement table_reset test in the new framework

bhuvana-talend:
# Description of change
https://jira.talendforge.org/browse/TDL-22960
- Rewrote this test from test_salesforce_incremental_table_reset.py to use the new framework. 
- It also provides the opportunity to test the new testcase table_reset_test.py in the new framework

# QA steps
 - [ ] automated tests passing
 
# Risks

# Rollback steps
 - revert this branch


tests/test_table_reset.py
```diff
@@ -0,0 +1,28 @@
+from sfbase import SFBaseTest
+from tap_tester.base_suite_tests.table_reset_test import TableResetTest
+
+
+class SFTableResetTest(TableResetTest, SFBaseTest):
+    """tap-salesforce Table reset test implementation"""
+
+    salesforce_api = 'BULK'
```
luandy64, line 8: This should probably go in `SFBaseTest` since it's used by `SFBaseTest.get_properties`

tests/test_table_reset.py
```diff
@@ -0,0 +1,28 @@
...
+    @staticmethod
+    def convert_state_to_utc(date_str):
+        """
+        Convert a saved bookmark value of form '2020-08-25T13:17:36-07:00' to
+        a string formatted utc datetime,
+        in order to compare aginast json formatted datetime values
+        """
+        date_object = dateutil.parser.parse(date_str)
+        date_object_utc = date_object.astimezone(tz=pytz.UTC)
+        return datetime.datetime.strftime(date_object_utc, "%Y-%m-%dT%H:%M:%SZ")
```
luandy64, line 28: This doesn't seem like a salesforce-table-reset specific function. Should we move this into the framework?

tests/test_table_reset.py
```diff
@@ -0,0 +1,28 @@
+from sfbase import SFBaseTest
+from tap_tester.base_suite_tests.table_reset_test import TableResetTest
+
+
+class SFTableResetTest(TableResetTest, SFBaseTest):
+    """tap-salesforce Table reset test implementation"""
+
+    salesforce_api = 'BULK'
+    reset_stream = 'User'
+    replication_method = "INCREMENTAL"
```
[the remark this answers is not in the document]
HarrisonMarcRose: This already exists in the tap-tester base_case. No need to define it again here.

```
    INCREMENTAL = "INCREMENTAL"  # replication method value in metadata
    FULL_TABLE = "FULL_TABLE"  # replication method value in metadata
```

tests/test_table_reset.py
HarrisonMarcRose: This is similar to parse_date and timedelta_formatted.  We should put this in base case in the same section or see if we can enhance that functionality to also convert to UTC.

tests/test_table_reset.py
HarrisonMarcRose: We should probably set a default in the base case, which can be overriden by setting the property here if needed.  Also, do we need to run this test for both BULK and REST?  We could import TableResetTest again in another module and class and set it in that test to REST.

bhuvana-talend: moved this to sfbase.py

tests/test_table_reset.py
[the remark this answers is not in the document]
bhuvana-talend: removed this

tests/test_table_reset.py
bhuvana-talend: moved this to framework base_case.py
This method in turn calls parse_date and then converts to utc

tests/sfbase.py
```diff
@@ -0,0 +1,956 @@
...
+        return stream in self.get_unsupported_by_rest_api()
+
+    def is_unsupported_by_bulk_api(self, stream):
+        """
+        returns True if stream is unsupported by BULK API
+
+        BULK API does not support any streams that are unsupported by the REST API and
+        in addition does not support the streams listed below.
+        """
+        return stream in self.get_unsupported_by_bulk_api()
```
HarrisonMarcRose, line 956: I thought we had discussed in slack that we were going to remove this as it isn't necessary yet and we have it in the other base and then we can add it in if needed when a test requires it as we transition to the new framework.

bhuvana-talend: yeah, missed it, Removed this method now

tests/sfbase.py
```diff
@@ -0,0 +1,943 @@
...
+            'WorkOrderLineItemStatus',
+            'ServiceAppointmentStatus',
+            'ServiceAppointmentStatus',
+            'FieldSecurityClassification',
+            # BUG_TODO | the following streams are undocumented
+            'WorkStepStatus',
+            'ShiftStatus',
+        }
+
+        return unsupported_streams_bulk_only | unsupported_streams_rest
```
HarrisonMarcRose, line 942: If we aren't using these methods in our new test (which it looks like we aren't) we should remove this.

tests/test_table_reset.py
```diff
@@ -0,0 +1,19 @@
+from sfbase import SFBaseTest
+from tap_tester.base_suite_tests.table_reset_test import TableResetTest
+
+
+class SFTableResetTest(TableResetTest, SFBaseTest):
+    """tap-salesforce Table reset test implementation
+    Currently tests only the stream with Incremental replication method"""
+
+    #reset_stream = 'User'
```
HarrisonMarcRose, line 9: Don't need this since we have it below.  Although, if you make this a property then this would work and you could remove the other one. I think that looks better but either way works.

HarrisonMarcRose (commented):
left a comment

tests/test_table_reset.py
bhuvana-talend: Made it as a property

HarrisonMarcRose (approved):

PR #162, opened 2023-07-05
Bookmark Bulk API 2.0 jobs so they can be reused on failure

alisonrclarke:
# Description of change

Adds bookmarks when creating Bulk API 2.0 jobs, so that if the sync fails or is cancelled, the results can be fetched from the same job on the next run.

If the JobID is no longer valid then the tap will restart the sync and create a new job.

# QA steps
 - [ ] automated tests passing
 
# Risks

# Rollback steps
 - revert this branch


alisonrclarke:
Closing as PR was opened in error.

PR #163, opened 2023-07-13
TDL-23556 Update PKs

luandy64:
# Description of change
This PR fixes the primary key for the `LightningUriEvent` Stream to make what Salesforce Docs says is the unique identifier of the event.

# QA steps
 - [x] automated tests passing
 - [x] manual qa steps passing (list below)
 
# Risks
- PK changes require a major version bump

# Rollback steps
 - revert this branch

### DIFF
--- a/tap_salesforce/__init__.py
+++ b/tap_salesforce/__init__.py
@@ -108,6 +108,9 @@ def create_property_schema(field, mdata):
 
     return (property_schema, mdata)
 
+PK_OVERRIDES = {
+    "LightningUriEvent": "EventIdentifier",
+}
 
 # pylint: disable=too-many-branches,too-many-statements
 def do_discover(sf):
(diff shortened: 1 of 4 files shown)


tap_salesforce/__init__.py
```diff
@@ -159,15 +161,16 @@ def do_discover(sf):
...
 
+        expected_pk_field = PK_OVERRIDES.get(sobject_name, "Id")
+
         # Loop over the object's fields
         for f in fields:
             field_name = f['name']
 
-            if field_name == "Id":
+            if field_name == expected_pk_field:
                 found_id_field = True
```
dsprayberry, line 171: I guess this doesn't really matter but I think we should rename this to be consistent with the new variable

dsprayberry (approved):
Had one comment but don't wanna block on it

dsprayberry (approved):

PR #165, opened 2023-08-04
Testcase to test the api quota limit

bhuvana-talend:
# Description of change
https://jira.talendforge.org/browse/TDL-23431
Run the sync job and make sure the per-run-limit is applied.
Split from feature https://jira.talendforge.org/browse/TDL-23349

# QA steps
 - [ ] automated tests passing
 
# Risks

# Rollback steps
 - revert this branch

### DIFF
--- a/tests/sfbase.py
+++ b/tests/sfbase.py
@@ -19,7 +19,7 @@ class SFBaseTest(BaseCase):
     """
 
     salesforce_api = "BULK"
-    total_quota = '80'
+    total_quota = '95'
     per_run_quota = None
     start_date = '2020-11-23T00:00:00Z'
 
--- a/tests/sfbase.py
+++ b/tests/sfbase.py
@@ -21,6 +21,7 @@ class SFBaseTest(BaseCase):
     salesforce_api = "BULK"
     total_quota = '95'
     per_run_quota = None
+    # default start date which can be overridden in the tests
     start_date = '2020-11-23T00:00:00Z'
 
     @staticmethod
@@ -33,9 +34,10 @@ class SFBaseTest(BaseCase):
         """the expected url route ending"""
         return "platform.salesforce"
 
-    def get_properties(self, original: bool = True):
+    def get_properties(self):
         """Configuration properties required for the tap."""
-        return_value = {
+
+        return {
             'start_date': self.start_date,
             'instance_url': 'https://singer2-dev-ed.my.salesforce.com',
             'select_fields_by_default': 'true',
@@ -45,21 +47,21 @@ class SFBaseTest(BaseCase):
             'is_sandbox': 'false'
         }
 
-        if original:
-            return return_value
-
-        # This test needs the new connections start date to be larger than the default
-        assert self.start_date > return_value["start_date"]
-
-        return_value["start_date"] = self.start_date
-        return return_value
-
     @staticmethod
     def get_credentials():
         """Authentication information for the test account"""
         return {'refresh_token': os.getenv('TAP_SALESFORCE_REFRESH_TOKEN'),
                 'client_id': os.getenv('TAP_SALESFORCE_CLIENT_ID'),
                 'client_secret': os.getenv('TAP_SALESFORCE_CLIENT_SECRET')}
+
+    @classmethod
+    def expected_stream_names(cls):
+        """A set of expected stream names"""
+        streams = set(cls.expected_metadata().keys())
+
+        if cls.salesforce_api == 'BULK':
+            return streams.difference(cls.rest_only_streams())
+        return streams
 
     @staticmethod
     def expected_metadata():
@@ -85,6 +87,12 @@ class SFBaseTest(BaseCase):
             BaseCase.REPLICATION_KEYS: {'LastModifiedDate'},
             BaseCase.REPLICATION_METHOD: BaseCase.INCREMENTAL,
         }
+
+        lightning_uri_event_full = {
+            BaseCase.PRIMARY_KEYS: {"EventIdentifier"},
+            BaseCase.REPLICATION_METHOD: BaseCase.FULL_TABLE,
+        }
+
         return {
             'AIApplication': default,  # removed # 6/13/2022 added back 7/10/2022
             'AIApplicationConfig': default,  # removed # 6/13/2022 added back 7/10/2022
@@ -449,7 +457,7 @@ class SFBaseTest(BaseCase):
             'LightningExperienceTheme': default,
             'LightningOnboardingConfig': default,
             'LightningToggleMetrics': default,  # new
-            'LightningUriEvent': default_full,  # new
+            'LightningUriEvent': lightning_uri_event_full,  # new
             'LightningUsageByAppTypeMetrics': default,  # new
             'LightningUsageByBrowserMetrics': default,  # new
             'LightningUsageByFlexiPageMetrics': default,  # new
@@ -751,17 +759,17 @@ class SFBaseTest(BaseCase):
             'WorkTypeGroupShare': incremental_last_modified,  # new
             'WorkTypeHistory': incremental_created_date,  # new
             'WorkTypeShare': incremental_last_modified,  # new
-            #'RecentlyViewed': default_full,  # new TODO verify this is not a bug  - Stream not discovered
-            #'TaskPriority': default,  # new TODO  - Stream not discovered
-            #'DeclinedEventRelation': default,  # new TODO  - Stream not discovered
-            #'AcceptedEventRelation': default,  # new TODO -  - Stream not discovered
-            #'OrderStatus': default,  # new TODO  - Stream not discovered
-            #'SolutionStatus': default,  # new TODO  - Stream not discovered
-            #'CaseStatus': default,  # new TODO  - Stream not discovered
-            #'TaskStatus': default,  # new TODO - Stream not discovered
-            # 'PartnerRole': default,  # new TODO - Stream not discovered
-            #'ContractStatus': default,  # new TODO  - Stream not discovered
-            #'UndecidedEventRelation': default,  # new TODO  - Stream not discovered
+            'RecentlyViewed': default_full,  # REST ONLY STREAM
+            'TaskPriority': default,  # REST ONLY STREAM
+            'DeclinedEventRelation': default,  # REST ONLY STREAM
+            'AcceptedEventRelation': default,  # REST ONLY STREAM
+            'OrderStatus': default,  # REST ONLY STREAM
+            'SolutionStatus': default,  # REST ONLY STREAM
+            'CaseStatus': default,  # REST ONLY STREAM
+            'TaskStatus': default,  # REST ONLY STREAM
+            'PartnerRole': default,  # REST ONLY STREAM
+            'ContractStatus': default,  # REST ONLY STREAM
+            'UndecidedEventRelation': default,  # REST ONLY STREAM
             # Newly discovered as of 2/12/2022
             'BriefcaseAssignment': default,
             'BriefcaseDefinition': default,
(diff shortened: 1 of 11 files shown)


tests/test_salesforce_api_per_run_quota.py
```diff
@@ -0,0 +1,53 @@
...
+            'quota_percent_total': '85',
+            'quota_percent_per_run' : '1',
+            'api_type': self.salesforce_api,
+            'is_sandbox': 'false'
+        }
+        if original:
+            return return_value
+
+        # This test needs the new connections start date to be larger than the default
+        assert self.start_date > return_value["start_date"]
```
bhtowles, line 40: What is the intent of this assertion?  I'm guessing that we just want to prevent start date from being too old.  If that is the case then I'd suggest setting start date dynamically.  If we are depending on what is specified in base, it will eventually have the potential to cause the same amount of issues as using the date defined in properties.

bhtowles (changes requested):

tests/test_salesforce_api_per_run_quota.py
```diff
@@ -0,0 +1,53 @@
+"""
+Test that with only custom fields selected for a stream automatic fields and custom fields  are still replicated
+"""
+import copy
```
bhtowles, line 4: copy does not appear to be used?

tests/test_salesforce_api_per_run_quota.py
```diff
@@ -0,0 +1,53 @@
...
+        # This test needs the new connections start date to be larger than the default
+        assert self.start_date > return_value["start_date"]
+
+        return_value["start_date"] = self.start_date
+        return return_value
+
+    def streams_to_test(self):
+        return self.expected_streams().difference( self.streams_to_exclude )
+
+    def test_all_streams_sync_records(self):
```
bhtowles, line 48: Test name does not seem accurate.  What are we really testing here?

tests/test_salesforce_api_per_run_quota.py
bhuvana-talend: Probably not needed, this method was overridden from sfbase.py

tests/sfbase.py
```diff
@@ -14,7 +14,14 @@
 
 class SFBaseTest(BaseCase):
 
+    """
+    Set these variables for the properties to use in the other tests
+    """
+
     salesforce_api = "BULK"
+    total_quota = '80'
```
bhtowles, line 22: both base and sfbase are moving to 95% total_quota

tests/sfbase.py
```diff
@@ -29,10 +36,11 @@ def get_type():
     def get_properties(self, original: bool = True):
         """Configuration properties required for the tap."""
         return_value = {
-            'start_date': '2020-11-23T00:00:00Z',
```
bhtowles, line 32: Looks like both our PRs are modifying get_properties.  To use the new pattern we have to return the return value and we are removing the original key word argument.

tests/sfbase.py
```diff
@@ -828,6 +836,14 @@ def expected_metadata():
             'WorkStepTemplateShare': incremental_last_modified,
         }
 
+    def expected_streams(self):
+        """A set of expected stream names"""
+        streams = set(self.expected_metadata().keys())
+
```
bhtowles, line 842: I implemented this the opposite way in my branch.  I should probably switch it back to match this since other other base file uses it too.  Let's discuss.

tests/test_salesforce_api_per_run_quota.py
```diff
@@ -0,0 +1,69 @@
...
+        return {}
+
+    def test_api_per_run_quota(self):
+        """
+        Run the tap in check mode and verify it returns the error for quota limit reached. If not, proceed to sync mode.
+        """
+        expected_per_run_error = "Terminating replication due to allotted quota"
+        conn_id = connections.ensure_connection(self)
+
+        found_catalogs = self.run_and_verify_check_mode( conn_id)
```
bhtowles, line 56: Does this handle the case where check mode fails?

tests/test_salesforce_api_per_run_quota.py
bhuvana-talend: Removed it

tests/sfbase.py
bhuvana-talend: changed it

tests/test_salesforce_api_per_run_quota.py
bhuvana-talend: I noticed that total quota check is caught in check mode and per run quota is caught in sync mode. I added that assertion as well to take care of the failure in check mode

bhtowles: Are we sure that total is always only caught in check mode and per run in sync mode?  Can we prove this and document it in the test?

tests/test_salesforce_api_per_run_quota.py
```diff
@@ -0,0 +1,75 @@
...
+        """
+        Run the tap in check mode and verify it returns the error for quota limit reached. If not, proceed to sync mode.
+        """
+        expected_per_run_error = "Terminating replication due to allotted quota"
+        expected_total_quota_error = "Terminating replication to not continue past configured percentage "
+        conn_id = connections.ensure_connection(self)
+
+        # run check mode
+        # Run a check job using orchestrator (discovery)
+        with self.assertRaises(Exception) as ex:
```
bhtowles, line 59: per run did not raise exception on last commit.  could the check job use less than 1 percent quota?  

```
FAIL: test_api_per_run_quota (test_salesforce_api_per_run_quota.SFAPIQuota)
Run the tap in check mode and verify it returns the error for quota limit reached. If not, proceed to sync mode.
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/root/project/tests/test_salesforce_api_per_run_quota.py", line 60, in test_api_per_run_quota
    found_catalogs = self.run_and_verify_check_mode( conn_id)
AssertionError: Exception not raised

2023-08-30 14:32:00,401 TapTester - INFO     ***** Test Results *****
----------------------------------------------------------------------
Ran 1 tests in 107s
```

bhtowles:
Also the table reset test is missing a manipulate state method.  Is this in another PR somewhere?  I thought I remembered someone talking about building that out.

tests/test_salesforce_api_per_run_quota.py
bhuvana-talend: Added the following comment in the test.
For the sync mode, we have a higher total quota set, so it is unlikely to hit the total quota. Noticed that per run quota limit reaches only during the sync mode.

bhuvana-talend:
> Also the table reset test is missing a manipulate state method. Is this in another PR somewhere? I thought I remembered someone talking about building that out.

https://jira.talendforge.org/browse/TDL-23972 - I also remember a conversation regarding this, created this card

tests/test_salesforce_api_per_run_quota.py
bhuvana-talend: We can't have less than 1%, couldn't send fraction amount. This is very tricky, let's discuss.

tests/test_salesforce_api_per_run_quota.py
```diff
@@ -0,0 +1,75 @@
...
+             'ShiftStatus',
+             'WorkStepStatus',
+        }
+
+    @staticmethod
+    def name():
+        return "tt_sf_api_quota"
+
+    def streams_to_test(self):
+        return self.expected_streams().difference( self.streams_to_exclude )
```
bhtowles, line 42:
```suggestion
        return self.expected_streams().difference(self.streams_to_exclude)
```

tests/test_salesforce_api_per_run_quota.py
```diff
@@ -0,0 +1,75 @@
...
+        Run the tap in check mode and verify it returns the error for quota limit reached. If not, proceed to sync mode.
+        For the sync mode, we have a higher total quota set, so it is unlikely to hit the total quota. Noticed that per run quota limit reaches only during the sync mode.
+        """
+        expected_per_run_error = "Terminating replication due to allotted quota"
+        expected_total_quota_error = "Terminating replication to not continue past configured percentage"
+        conn_id = connections.ensure_connection(self)
+
+        # Run a check job using orchestrator (discovery)
+        with self.assertRaises(Exception) as ex:
+             found_catalogs = self.run_and_verify_check_mode( conn_id)
```
bhtowles, line 60:
```suggestion
             found_catalogs = self.run_and_verify_check_mode(conn_id)
```

tests/test_salesforce_api_per_run_quota.py
```diff
@@ -0,0 +1,75 @@
...
+
+        self.assertIn(expected_total_quota_error, str(ex.exception) )
```
bhtowles, line 62:
```suggestion
        self.assertIn(expected_total_quota_error, str(ex.exception))
```

tests/test_salesforce_api_per_run_quota.py
```diff
@@ -0,0 +1,75 @@
...
+        test_catalogs = [catalog for catalog in found_catalogs
+                           if catalog.get('stream_name') in self.streams_to_test()]
+
+        # non_selected_fields are none
+        self.perform_and_verify_table_and_field_selection(conn_id, test_catalogs)
+
+        with self.assertRaises(Exception) as ex:
+             record_count_by_stream = self.run_and_verify_sync_mode(conn_id)
+
+        self.assertIn(expected_per_run_error, str(ex.exception) )
```
bhtowles, line 74:
```suggestion
        self.assertIn(expected_per_run_error, str(ex.exception))
```

tests/test_salesforce_api_total_quota.py
```diff
@@ -0,0 +1,57 @@
...
+             'ShiftStatus',
+             'WorkStepStatus',
+        }
+
+    @staticmethod
+    def name():
+        return "tt_sf_api_quota_total"
+
+    def streams_to_test(self):
+        return self.expected_streams().difference( self.streams_to_exclude )
```
bhtowles, line 41:
```suggestion
        return self.expected_streams().difference(self.streams_to_exclude)
```

tests/test_salesforce_api_total_quota.py
```diff
@@ -0,0 +1,57 @@
...
+        expected_total_quota_error = "Terminating replication to not continue past configured percentage of 1.0% total quota"
+
+        conn_id = connections.ensure_connection(self)
+
+        # run check mode
+        # Run a check job using orchestrator (discovery)
+        with self.assertRaises(Exception) as ex:
+             check_job_name = self.run_and_verify_check_mode( conn_id)
+
+        self.assertIn(expected_total_quota_error, str(ex.exception) )
```
bhtowles, line 56:
```suggestion
        self.assertIn(expected_total_quota_error, str(ex.exception))
```

tests/test_salesforce_api_per_run_quota.py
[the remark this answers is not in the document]
bhuvana-talend: removed

tests/test_salesforce_api_total_quota.py
bhuvana-talend: sorry, missed some comments. I removed some more white spaces and hopefully I took care of all of them

tests/test_salesforce_api_total_quota.py
```diff
@@ -0,0 +1,57 @@
...
+
+    def test_api_total_quota(self):
+        """
+        Run the tap in check mode and verify it returns the error for quota limit reached. 
+        """
+        expected_total_quota_error = "Terminating replication to not continue past configured percentage of 1.0% total quota"
+
+        conn_id = connections.ensure_connection(self)
+
+        # run check mode
```
bhtowles, line 51:
```suggestion
```

tests/test_salesforce_api_per_run_quota.py
bhtowles: If it is not guaranteed to hit the per run quota in check mode can we do a "try / catch" for check mode and then use the normal "assertRaises" for sync mode?  That way if the quota is hit during check mode we can catch the error and verify it and if it is not hit until sync mode we can also catch it there.  If that works I'd use the same pattern for both tests.

bhtowles (changes requested):

PR #166, opened 2023-08-04
Add custom object to saleseforce account and update base to match

bhtowles:
# Description of change
Add custom objects to expected metadata

# QA steps
 - [ ] automated tests passing
 
# Risks

# Rollback steps
 - revert this branch

### DIFF
--- a/tests/base.py
+++ b/tests/base.py
@@ -822,6 +822,8 @@ class SalesforceBaseTest(BaseCase):
             'ShiftHistory': incremental_created_date,
             'ShiftShare': incremental_last_modified,
             'ShiftStatus': default,
+            'TapTester__c': default, # added 8/4/2023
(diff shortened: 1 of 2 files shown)


HarrisonMarcRose (commented):
Curious what tests this would be picked up in?  I assume discovery automatically, but are there any others or is that being done separately?

bhtowles:
> Curious what tests this would be picked up in? I assume discovery automatically, but are there any others or is that being done separately?
> 
> Also, it looks like two tests are failing in discovery for some reason.

From the build_daily it looks like discovery rest, discovery bulk, and table reset.  Not sure why table reset is picking this up automatically but there are new edits / tests planned to explicitly pick up these custom streams before closing this card.  This first merge is just to fix the existing tests.  There will be a separate PR for the new / modified tests.

PR #167, opened 2023-08-09
Verify tap can replicate all fields from new custom object stream

bhtowles:
# Description of change
TDL-17114

# QA steps
 - [ ] automated tests passing
 
# Risks

# Rollback steps
 - revert this branch

### DIFF
--- a/tests/sfbase.py
+++ b/tests/sfbase.py
@@ -78,6 +78,12 @@ class SFBaseTest(BaseCase):
             BaseCase.REPLICATION_KEYS: {'LastModifiedDate'},
             BaseCase.REPLICATION_METHOD: BaseCase.INCREMENTAL,
         }
+
+        lightning_uri_event_full = {
+            self.PRIMARY_KEYS: {"EventIdentifier"},
+            self.REPLICATION_METHOD: self.FULL_TABLE,
+        }
+
         return {
             'AIApplication': default,  # removed # 6/13/2022 added back 7/10/2022
             'AIApplicationConfig': default,  # removed # 6/13/2022 added back 7/10/2022
@@ -442,7 +448,7 @@ class SFBaseTest(BaseCase):
             'LightningExperienceTheme': default,
             'LightningOnboardingConfig': default,
             'LightningToggleMetrics': default,  # new
-            'LightningUriEvent': default_full,  # new
+            'LightningUriEvent': lightning_uri_event_full,  # new
             'LightningUsageByAppTypeMetrics': default,  # new
             'LightningUsageByBrowserMetrics': default,  # new
             'LightningUsageByFlexiPageMetrics': default,  # new
--- a/tests/sfbase.py
+++ b/tests/sfbase.py
@@ -80,8 +80,8 @@ class SFBaseTest(BaseCase):
         }
 
         lightning_uri_event_full = {
-            self.PRIMARY_KEYS: {"EventIdentifier"},
-            self.REPLICATION_METHOD: self.FULL_TABLE,
+            BaseCase.PRIMARY_KEYS: {"EventIdentifier"},
+            BaseCase.REPLICATION_METHOD: self.FULL_TABLE,
         }
 
         return {
--- a/tests/sfbase.py
+++ b/tests/sfbase.py
@@ -801,6 +801,8 @@ class SFBaseTest(BaseCase):
             'ShiftHistory': incremental_created_date,
             'ShiftShare': incremental_last_modified,
             'ShiftStatus': default,
+            'TapTester__c': default, # added 8/4/2023
+            'TapTester__Share': incremental_last_modified, # added 8/4/2023
             'WebCart': default,  # re-added # 10/18/2022
             'WebCartAdjustmentGroup': default,  # added # 10/18/2022
             'WebCartHistory': incremental_created_date,  # re-added # 10/18/2022
--- a/tests/sfbase.py
+++ b/tests/sfbase.py
@@ -49,9 +49,9 @@ class SFBaseTest(BaseCase):
     @classmethod
     def expected_stream_names(cls):
         """A set of expected stream names"""
-        streams = set(self.expected_metadata().keys())
+        streams = set(cls.expected_metadata().keys())
 
-        if self.salesforce_api == 'BULK':
+        if cls.salesforce_api == 'BULK':
             return streams.difference(cls.rest_only_streams())
         return streams
 
--- a/tests/sfbase.py
+++ b/tests/sfbase.py
@@ -15,6 +15,8 @@ from tap_tester.base_suite_tests.base_case import BaseCase
 class SFBaseTest(BaseCase):
 
     salesforce_api = "BULK"
+    # default start date which can be overridden in the tests
+    start_date = '2020-11-23T00:00:00Z'
 
     @staticmethod
     def tap_name():
@@ -26,10 +28,10 @@ class SFBaseTest(BaseCase):
         """the expected url route ending"""
         return "platform.salesforce"
 
-    def get_properties(self, original: bool = True):
+    def get_properties(self):
         """Configuration properties required for the tap."""
-        return_value = {
-            'start_date': '2020-11-23T00:00:00Z',
+        return {
+            'start_date': self.start_date,
             'instance_url': 'https://singer2-dev-ed.my.salesforce.com',
             'select_fields_by_default': 'true',
             'quota_percent_total': '95',
@@ -37,21 +39,20 @@ class SFBaseTest(BaseCase):
             'is_sandbox': 'false'
         }
 
-        if original:
-            return return_value
-
-        # This test needs the new connections start date to be larger than the default
-        assert self.start_date > return_value["start_date"]
-
-        return_value["start_date"] = self.start_date
-        return return_value
-
     @staticmethod
     def get_credentials():
         """Authentication information for the test account"""
         return {'refresh_token': os.getenv('TAP_SALESFORCE_REFRESH_TOKEN'),
                 'client_id': os.getenv('TAP_SALESFORCE_CLIENT_ID'),
                 'client_secret': os.getenv('TAP_SALESFORCE_CLIENT_SECRET')}
+
+    @classmethod
+    def expected_stream_names(cls):
+        """A set of expected stream names"""
+        if cls.salesforce_api == 'BULK':
+            return set(cls.expected_metadata().keys())
+        else:
+            return set(cls.expected_metadata().keys()).union(cls.rest_only_streams())
 
     @staticmethod
     def expected_metadata():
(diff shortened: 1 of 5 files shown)


bhuvana-talend:
Are these 2 new files ? 
For tap-salesforce, any new tests we wanted to add using the new framework. Is there any specific reason for using the old framework and not implementing with new framework ?

bhuvana-talend:
Changes look good to me

bhuvana-talend (approved):

bhtowles:
> Are these 2 new files ?
> For tap-salesforce, any new tests we wanted to add using the new framework. Is there any specific reason for using the old framework and not implementing with new framework ?

Yes, 2 new files.  This was kept out of the base suite since custom objects are not supported across most Tier 1 taps.   Also the naming conventions (__c for example) may be different from tap to tap.  It is certainly worth further discussion to make sure there isn't a way we could stream line this to make it more of a one size fits all approach.

tests/test_custom_objects_rest.py
```diff
@@ -0,0 +1,100 @@
+"""
+Test that all fields can be replicated for a stream that is a custom object (REST API)
+"""
+from datetime import datetime, timedelta
+
+from tap_tester import connections, menagerie, runner
+
+from base import SalesforceBaseTest
```
HarrisonMarcRose, line 8: is this the new base or the old one.  We should be writing all tests against the new one.

tests/test_custom_objects_rest.py
```diff
@@ -0,0 +1,100 @@
...
+    """Test that all fields can be replicated for a stream that is a custom object"""
+
+    # Note: custom record as seen Aug 9, 2023
+    # TODO do we need to hard code start date to 08-04?
+    # 'CreatedDate': '2023-08-04T20:13:50.000000Z'
+    # 'SystemModstamp': '2023-08-04T20:13:50.000000Z',
+    # 'LastModifiedDate': '2023-08-04T20:13:50.000000Z',
+    # 'LastReferencedDate': '2023-08-09T20:19:45.000000Z',
+    # 'LastViewedDate': '2023-08-09T20:19:45.000000Z',
+    start_date =  (datetime.now() + timedelta(days=-1)).strftime("%Y-%m-%dT00:00:00Z")
```
HarrisonMarcRose, line 21: Your TODO and the dates of these records look like yesterday won't work for long.  Is this a static record where we need a hardcoded start date or are we okay?

tests/test_custom_objects_rest.py
```diff
@@ -0,0 +1,100 @@
...
+        Verify that all fields can be replicated for a custom  stream
+
+        PREREQUISITE
+        Create a custom object and at least one data record for that cusotm object
+        """
+
+        expected_streams = self.expected_sync_streams()
+
+        # instantiate connection
+        conn_id = connections.ensure_connection(self, original_properties=False)
```
HarrisonMarcRose, line 41: In the new framework we generally aren't using original properties any longer as this is to change start date which we made a property we can easily override instead.

tests/test_custom_objects_rest.py
```diff
@@ -0,0 +1,100 @@
...
+        for stream in expected_streams:
+            with self.subTest(stream=stream):
+
+                # expected values
+                expected_keys = stream_to_all_catalog_fields.get(stream)
+
+                # collect actual values
+                data = synced_records.get(stream)
+                record_messages_keys = [set(row['data'].keys()) for row in data['messages']
+                                        if row['action'] == 'upsert']
```
HarrisonMarcRose, line 78: hmmm. I'm not sure I understand this.  Looks like you are going to end up with a list of multiple sets of keys.  I would think we want a single list of all the keys in the messages.  Maybe you are planning doing a union of the members of the list next?

tests/test_custom_objects_rest.py
```diff
@@ -0,0 +1,100 @@
...
+
+                # Verify that you get some records for each stream
+                self.assertGreater(
+                    record_count_by_stream.get(stream, -1), 0,
+                    msg="The number of records is not over the stream max limit")
+
+                # Verify that only the automatic fields are sent to the target
+                for actual_keys in record_messages_keys:
+                    self.assertSetEqual(expected_keys, actual_keys)
```
HarrisonMarcRose, line 87: Ah. So we are expecting every record to have all the keys.  I guess this works if we either have data for every record or if we still send keys for null fields.

tests/test_custom_objects_rest.py
bhtowles: I don't know.  The reason all these dates are still commented there was the initial effort to understand why the record is still being replicated since the start date is now well past the static record creation time.  I was going to change the start date but since it wasn't technically broken I didn't fix it.

tests/test_custom_objects_rest.py
bhtowles: Will update to new base file.

tests/test_custom_objects_rest.py
bhtowles: Will remove along with update to new base file.

tests/test_custom_objects_rest.py
bhtowles: Related to below comment.

tests/test_custom_objects_rest.py
bhtowles: Was expecting data for every record but I'm not sure about null fields.  Does this need an update to support that?

tests/test_custom_objects_rest.py
bhtowles: Follow up card to be written to determine what the expected behavior is for custom objects with no defined replication method.  Determine how to set replication method and update test to expect correct behavior.

bhtowles: https://jira.talendforge.org/browse/TDL-23886

tests/test_custom_objects_rest.py
luandy64, line 1: The file structure here is strange.

Why is `SalesforceCustomObjects` in the `test_custom_objects_rest.py` file?
   - Because of this, `test_custom_objects_bulk.py` depends on `test_custom_objects_rest.py`, but there's no dependency between `bulk` and `rest`, they are siblings

tests/test_custom_objects_rest.py
```diff
@@ -0,0 +1,104 @@
...
+
+        # instantiate connection
+        conn_id = connections.ensure_connection(self)
+
+        # run check mode
+        found_catalogs = self.run_and_verify_check_mode(conn_id)
+
+        # table and field selection
+        test_catalogs = [catalog for catalog in found_catalogs
+                                          if catalog.get('stream_name') in expected_streams]
```
luandy64, line 48: Style Note: The second line of a longer line usually lines up with second character of the first line.

```python
result = [{'key': value} for value in iterable
          if a_long_filter_expression(value)]
```
[Source](https://google.github.io/styleguide/pyguide.html#274-decision)

---

My personal list comprehension rule is that if it doesn't fit on one line, each segment gets its own line

```python
test_catalogs = [
    catalog 
    for catalog in found_catalogs
    if catalog.get('stream_name') in expected_streams
]
```

tests/test_custom_objects_rest.py
```diff
@@ -0,0 +1,104 @@
...
+
+        self.perform_and_verify_table_and_field_selection(conn_id, test_catalogs)
+
+        # grab metadata after performing table-and-field selection to set expectations
+        #   used for asserting all fields are replicated
+        stream_to_all_catalog_fields = dict()
+        for catalog in test_catalogs:
+            stream_id, stream_name = catalog['stream_id'], catalog['stream_name']
```
luandy64, line 56: These are used once each, maybe consider not binding them to a variable

tests/test_custom_objects_rest.py
```diff
@@ -0,0 +1,104 @@
...
+            catalog_entry = menagerie.get_annotated_schema(conn_id, stream_id)
+            fields_from_field_level_md = [md_entry['breadcrumb'][1]
+                                          for md_entry in catalog_entry['metadata']
+                                          if md_entry['breadcrumb'] != []]
```
luandy64, line 60: This could be a set comprehension since the next line immediately turns this list into a set

tests/test_custom_objects_rest.py
```diff
@@ -0,0 +1,104 @@
...
+            'TapTester__c',  # 1 created record present, spike on CRUD?
+            # 'TapTester__Share',  # TODO no data, fix or omit?
+        }
+
+    def custom_objects_test(self):
+        """
+        Verify that all fields can be replicated for a custom  stream
+
+        PREREQUISITE
+        Create a custom object and at least one data record for that cusotm object
```
luandy64, line 35:
```suggestion
        Create a custom object and at least one data record for that custom object
```

tests/test_custom_objects_rest.py
```diff
@@ -0,0 +1,104 @@
...
+    @staticmethod
+    def expected_sync_streams():
+        return {
+            'TapTester__c',  # 1 created record present, spike on CRUD?
+            # 'TapTester__Share',  # TODO no data, fix or omit?
+        }
+
+    def custom_objects_test(self):
+        """
+        Verify that all fields can be replicated for a custom  stream
```
luandy64, line 32:
```suggestion
        Verify that all fields can be replicated for a custom stream
```

tests/test_custom_objects_rest.py
```diff
@@ -0,0 +1,104 @@
...
+                data = synced_records.get(stream)
+                record_messages_keys = [set(row['data'].keys()) for row in data['messages']
+                                        if row['action'] == 'upsert']
+
+                # Verify that you get some records for each stream
+                self.assertGreater(
+                    record_count_by_stream.get(stream, -1), 0,
+                    msg="The number of records is not over the stream max limit")
+
+                # Verify that only the automatic fields are sent to the target
```
luandy64, line 83: This comment doesn't seem to agree with the rest of the file

Test description:
> Verify that all fields can be replicated for a custom  stream

The expected value of the assert this comment is about
> expected_keys = stream_to_all_catalog_fields.get(stream)

luandy64: Though it is confusing that both the bulk test and the rest test override this

```python
    @staticmethod
    def streams_to_selected_fields():
        """Note: if this is overridden you are not selecting all fields.
        Therefore this should rarely if ever be used for this test."""
        return {}
```

So then are the two examples I pulled out (`Verify that all fields can be replicated for a custom stream` and `expected_keys = stream_to_all_catalog_fields.get(stream)`) actually the problem?

tests/sfbase.py
```diff
@@ -743,17 +751,17 @@ def expected_metadata():
...
+            'TaskPriority': default,  # REST ONLY STREAM
+            'DeclinedEventRelation': default,  # REST ONLY STREAM
+            'AcceptedEventRelation': default,  # REST ONLY STREAM
+            'OrderStatus': default,  # REST ONLY STREAM
+            'SolutionStatus': default,  # REST ONLY STREAM
+            'CaseStatus': default,  # REST ONLY STREAM
+            'TaskStatus': default,  # REST ONLY STREAM
+            'PartnerRole': default,  # REST ONLY STREAM
+            'ContractStatus': default,  # REST ONLY STREAM
+            'UndecidedEventRelation': default,  # REST ONLY STREAM
```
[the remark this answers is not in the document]
bhtowles: Just to be more transparent.  These streams were originally commented out of sfbase but they were left in the base file and I wasn't sure why.  I later discovered the logic and dependancies.  base left them in so that a list of all keys could be created from iterating over expected_metadata.  These streams are removed from the list by other methods to account for bulk api usage.  I figured it may be more clear to label them in both places (expected_metadata and the rest only streams function) to make it more clear going forward.

tests/test_custom_objects_rest.py
bhtowles: Good point.  Was just following the convention of the discovery test files.  I'll bring this up with the team but not address it now.

tests/test_custom_objects_rest.py
bhtowles: Updated

tests/test_custom_objects_rest.py
bhtowles: Updated

tests/test_custom_objects_rest.py
bhtowles: Updated

tests/test_custom_objects_rest.py
bhtowles: Updated comment.  Yea it's a little confusing the way stream and field selection is set up in the framework.

PR #173, opened 2023-11-08
Switch replication methods

JYOTHINARAYANSETTY:
# Description of change
Switching replication method from full_table to incremental and vice versa

# QA steps
 - [ ] automated tests passing
 
# Risks
low
# Rollback steps
 - revert this branch

### DIFF
--- a/tests/test_salesforce_switch_rep_method_ft_incrmntl.py
+++ b/tests/test_salesforce_switch_rep_method_ft_incrmntl.py
@@ -84,6 +84,10 @@ class SFSwitchRepMethodIncrmntl(SFBaseTest):
                 #Verify all records are synced in the second sync
                 self.assertTrue(fulltbl_primary_keys.issubset(incrmntl_primary_keys))
 
+                """
+                Modify the the activate version assertion accordingly based on the outcome of BUG #TDL-24467
+                if needed
+                """
                 #verify that the last message is not a activate version message for incremental sync
                 self.assertNotEqual('activate_version', incrmntl_sync_records[stream]['messages'][-1]['action'])
 
--- a/tests/test_salesforce_switch_rep_method_ft_incrmntl.py
+++ b/tests/test_salesforce_switch_rep_method_ft_incrmntl.py
@@ -12,6 +12,7 @@ class SFSwitchRepMethodIncrmntl(SFBaseTest):
 
     def expected_sync_streams(self):
         streams = self.switchable_streams() - {'FlowDefinitionView','EntityDefinition'}
+        # Excluded the above two streams due to the bug TDL-24514
         return self.partition_streams(streams)
 
 
@@ -62,11 +63,19 @@ class SFSwitchRepMethodIncrmntl(SFBaseTest):
                 self.assertGreater(incrmntl_sync_count, 0,
                                    msg="We are not fully testing bookmarking for {}".format(stream))
                 # data from record messages
+                """
+                If implementing in tap-tester framework the primary key implementation should account
+                for compound primary keys
+                """
+                self.assertEqual(1, len(list(primary_keys[stream])),
+                                 msg="Compound primary keys require a change to test expectations")
                 primary_key = list(primary_keys[stream])[0]
                 fulltbl_sync_messages = [record['data'] for record in
                                        fulltbl_sync_records.get(stream).get('messages')
                                        if record.get('action') == 'upsert']
-                fulltbl_primary_keys = {message[primary_key] for message in fulltbl_sync_messages}
+                filtered_fulltbl_sync_messages = [message for message in fulltbl_sync_messages
+                                                  if message[replication_key] >= self.start_date]
+                fulltbl_primary_keys = {message[primary_key] for message in filtered_fulltbl_sync_messages}
                 incrmntl_sync_messages = [record['data'] for record in
                                         incrmntl_sync_records.get(stream).get('messages')
                                         if record.get('action') == 'upsert']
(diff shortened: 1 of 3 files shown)


tests/test_salesforce_switch_rep_method_ft_incrmntl.py
```diff
@@ -0,0 +1,96 @@
...
+                fulltbl_primary_keys = {message[primary_key] for message in fulltbl_sync_messages}
+                incrmntl_sync_messages = [record['data'] for record in
+                                        incrmntl_sync_records.get(stream).get('messages')
+                                        if record.get('action') == 'upsert']
+                incrmntl_primary_keys = {message[primary_key] for message in incrmntl_sync_messages}
+
+                #Verify all records are synced in the second sync
+                self.assertTrue(fulltbl_primary_keys.issubset(incrmntl_primary_keys))
+
+                #verify that the last message is not a activateversion message for incremental sync
```
bhtowles, line 78: What was the final decision on the leading 'action'?  I'd like to also verify that the leading 'action' is either activate_version or upsert as expected.

tests/test_salesforce_switch_rep_method_incrmntl_ft.py
```diff
@@ -0,0 +1,96 @@
...
+                fulltbl_primary_keys = {message[primary_key] for message in fulltbl_sync_messages}
+
+                #Verify all records are synced in the second sync
+                self.assertTrue(incrmntl_primary_keys.issubset(fulltbl_primary_keys))
+
+                #Verify that the fulltable sync count is greater or equal to incrmental sync count
+                self.assertGreaterEqual(fulltbl_sync_count, incrmntl_sync_count,
+                                        msg = "Full table sync didn't fetch all the records")
+
+                #verify that last messages of every stream is the activate version message
```
bhtowles, line 79: same comment on leading activate versions

tests/test_salesforce_switch_rep_method_ft_incrmntl.py
```diff
@@ -0,0 +1,96 @@
...
+                                       fulltbl_sync_records.get(stream).get('messages')
+                                       if record.get('action') == 'upsert']
+                fulltbl_primary_keys = {message[primary_key] for message in fulltbl_sync_messages}
+                incrmntl_sync_messages = [record['data'] for record in
+                                        incrmntl_sync_records.get(stream).get('messages')
+                                        if record.get('action') == 'upsert']
+                incrmntl_primary_keys = {message[primary_key] for message in incrmntl_sync_messages}
+
+                #Verify all records are synced in the second sync
+                self.assertTrue(fulltbl_primary_keys.issubset(incrmntl_primary_keys))
```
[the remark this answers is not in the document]
JYOTHINARAYANSETTY: In this scenario, a few records are updated on each sync, specifically the replication key value, thus when we filter the incremental sync, we are actually removing records that were generated previously but were recently altered.
As a result, the equal primary key sets assumption fails. 
After talking with Harrison, we concluded that it should be fine if the full_table sync primary keys are a subset of the Incremental sync primary keys in case any new entries are added after the full_table sync.

tests/test_salesforce_switch_rep_method_ft_incrmntl.py
JYOTHINARAYANSETTY: This would depend on the ticket [TDL-24467](https://jira.talendforge.org/browse/TDL-24467) Will add  a comment for time being  to add/modify assertion based on the ticket implementation

tests/test_salesforce_switch_rep_method_incrmntl_ft.py
JYOTHINARAYANSETTY: This would depend on the ticket [TDL-24467](https://jira.talendforge.org/browse/TDL-24467) Will add a comment for time being to add/modify assertion based on the ticket implementation

tests/test_salesforce_switch_rep_method_ft_incrmntl.py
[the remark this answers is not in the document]
HarrisonMarcRose: Correct. This assures that the incremental sync has all the records from the original full table sync, ignoring extra records.  

I'm not sure what the answer was for salesforce about full table adhering to the start date or ignoring the start date.  This wouldn't work if the full table sync ignored the start date and went before it.  This could be okay with an early enough start date, but when we put this in tap tester, we should generalize this to filter out records before the start date in the full sync records.

tests/test_salesforce_switch_rep_method_ft_incrmntl.py
```diff
@@ -0,0 +1,96 @@
...
+
+                #verify that the last message is not a activateversion message for incremental sync
+                self.assertNotEqual('activate_version', incrmntl_sync_records[stream]['messages'][-1]['action'])
```
HarrisonMarcRose, line 79: The activate_version message tells the loaders it is okay to go ahead and load data into the table. If we have sent all of the data, it is okay to start loading.  So although this message may not be necessary, it would have no negative consequences.

Also, If they take the appropriate action and fix it so that they do not send the activate_version as the first message for a new table version for incremental, waiting until all data is ready before deleting the previous full table records, Then it would be necessary/critical to have this message be sent.

So in the final version we should assert that the only the last message is an activate version, and as a workaround we should verify that there is an activate message being sent.

tests/test_salesforce_switch_rep_method_incrmntl_ft.py
```diff
@@ -0,0 +1,96 @@
...
+                self.assertEqual('activate_version', fulltbl_sync_records[stream]['messages'][fulltbl_sync_count+1]
+                                 ['action'])
```
HarrisonMarcRose, line 81: Also, the same comment as above. I think we should expect this and also when the ticket is fixed assert that there is no other message that is an activate_version message.  We should probably write out that assertion that will fail and comment it out with a comment mentioning the ticket that will address the issue.

tests/test_salesforce_switch_rep_method_ft_incrmntl.py
```diff
@@ -0,0 +1,96 @@
...
+class SFSwitchRepMethodIncrmntl(SFBaseTest):
+
+    start_date = '2000-11-11T00:00:00Z'#to get max data available for testing
+    @staticmethod
+    def name():
+        return "tt_sf_table_switch_rep_method_ft_incrmntl"
+
+
+    def expected_sync_streams(self):
+        streams = self.switchable_streams() - {'FlowDefinitionView','EntityDefinition'}
```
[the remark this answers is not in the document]
JYOTHINARAYANSETTY: The records data for these streams is weird , there is no value against replication key for most of the records and the primary key is same for all. Incremental sync is mostly returning a empty set for these streams.

tests/test_salesforce_switch_rep_method_ft_incrmntl.py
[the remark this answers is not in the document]
JYOTHINARAYANSETTY: Full table doesn't adhere to start date. Will modify to filter the full table sync records according to the start date.

tests/test_salesforce_switch_rep_method_ft_incrmntl.py
JYOTHINARAYANSETTY: Added a qa task in the DOD of the ticket to modify the assertions based on the final implementation.

tests/test_salesforce_switch_rep_method_incrmntl_ft.py
HarrisonMarcRose: I don't see a change here.

HarrisonMarcRose (approved):
Approving, but I don't see a change or comment for the activate version in the second test

tests/test_salesforce_switch_rep_method_incrmntl_ft.py
JYOTHINARAYANSETTY: For the activate version, added a qa task/note in the ticket itself in the DOD to add or m,odify the assertions based on the final implementation.

tests/test_salesforce_switch_rep_method_ft_incrmntl.py
```diff
@@ -0,0 +1,105 @@
...
+                #Verify all records are synced in the second sync
+                self.assertTrue(fulltbl_primary_keys.issubset(incrmntl_primary_keys))
+
+                #verify that the last message is not a activate version message for incremental sync
+                self.assertNotEqual('activate_version', incrmntl_sync_records[stream]['messages'][-1]['action'])
+
+                #verify that the table version incremented after every sync
+                self.assertGreater(incrmntl_sync_records[stream]['table_version'],
+                                    fulltbl_sync_records[stream]['table_version'],
+                                   msg = "Table version is not incremented after a successful sync")
```
bhtowles, line 93: Indentation

PR #218, opened 2026-05-07
SAC-30747: use composite batch API for discovery

skuttleman:
# Description of change
https://qlik-dev.atlassian.net/browse/SAC-30747
Refactor discovery code to use [composite batchAPI]() to speed up discovery jobs

# QA steps
 - [x] automated tests passing
 - [x] manual qa steps passing (list below)
   - compared schema output from current code vs new code -> same (but in different order)
 
# Risks

# Rollback steps
 - revert this branch

#### AI generated code
https://internal.qlik.dev/general/ways-of-working/code-reviews/#guidelines-for-ai-generated-code
- [x] this PR has been written with the help of GitHub Copilot or another generative AI tool

### DIFF
--- a/tap_salesforce/salesforce/__init__.py
+++ b/tap_salesforce/salesforce/__init__.py
@@ -431,7 +431,9 @@ class Salesforce():
             if resp_json["hasErrors"]:
                 errors = [result["result"] for result in resp_json["results"] if result["statusCode"] != 200]
                 raise TapSalesforceException(f"Error(s) contained in composite batch response: {errors}")
-
+# hasErrors was false but result count is still wrong
+if len(resp_json["results"]) != len(chunk):
+    raise TapSalesforceException(f"Composite batch returned {len(resp_json['results'])} results but expected {len(chunk)}")
             for name, result in zip(chunk, resp_json["results"]):
                 results[name] = result["result"]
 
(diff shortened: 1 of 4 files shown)


KrisPersonal (commented):
Can the batch describe size be more than 25?

skuttleman:
> Can the batch describe size be more than 25?

No. `25` is the max: https://developer.salesforce.com/docs/atlas.en-us.api_rest.meta/api_rest/resources_composite_batch.htm

humbamp123 (approved):

tap_salesforce/salesforce/__init__.py
```diff
@@ -385,24 +386,57 @@ def login(self):
...
+            with metrics.http_request_timer("describe") as timer:
+                timer.tags['endpoint'] = "composite_batch"
+                resp = self._make_request('POST', url, headers=headers, body=json.dumps(body))
+
+            resp_json = resp.json()
+
+            if resp_json["hasErrors"]:
+                errors = [result["result"] for result in resp_json["results"] if result["statusCode"] != 200]
+                raise TapSalesforceException(f"Error(s) contained in composite batch response: {errors}")
+
```
leslievandemark, line 434: From copilot: 
The new `batch_describe` method maps Salesforce results back to object names using `zip(chunk, resp_json["results"])`. Python's zip silently stops at the shortest iterable, so if Salesforce returns fewer results than requested (even with `hasErrors: False`), the remaining objects are dropped from discovery with no error raised. The old per-object describe would have failed loudly in this case. This guard ensures a count mismatch raises an explicit exception rather than silently producing incomplete schema output. 
```suggestion
# hasErrors was false but result count is still wrong
if len(resp_json["results"]) != len(chunk):
    raise TapSalesforceException(f"Composite batch returned {len(resp_json['results'])} results but expected {len(chunk)}")
```
We shouldn't need this if the salesforce api works as expected, but wouldn't hurt to add.

leslievandemark (approved):

